Issue #16, opened 2016-02-02
Tabs component throws error when using links as `src` attribute


Stanton:
![href_360 gif png](https://cloud.githubusercontent.com/assets/18653/12752677/9096873a-c9bb-11e5-8c5e-022e13a2d215.gif)

PR #17, opened 2016-02-03
Tabs using links as the `src` should no longer throw a javascript error

Stanton:
Modified tabs helper to no longer pass `data-toggle="tab"` when using full or partial URLs


Stanton:
Resolves #16

PR #43, opened 2016-02-17
[PUL-117] Possible fix for form.end spacing

james-manley:


Stanton:
Addresses #44

views/pulsar/v2/helpers/form.html.twig
```diff
@@ -573,7 +573,6 @@ class   | string | A space separated list of class names
 
 #}
 {% macro end(options = {}) %}
-{% spaceless %}
```
Stanton: Wouldn't've thought removing the top level `spaceless` was required? Didn't removing the whitespace control on L:588 resolve the issue alone?

james-manley: I'd have thought so too, but seems on a combination of removing both solves it. I'm no expert in Twig stuff though!

Stanton: Yep, you're right!

PR #50, opened 2016-02-18
Add a test case for a reported issue with using both `form` and `html` macros within the `form.group` helper

Stanton:
Test passes fine


langy:
:+1:

PR #55, opened 2016-02-19
Only use `inline-block` on button groups that aren't being used as form controls.

### LINKED ISSUE #54: Button groups aren't vertically aligned with other form controls
Due to button groups having `display: inline-block;` applied, they misalign with other form controls when used on a form that uses the `form__group` elements, whose children are styled using `display: table-cell;`
screen shot 2016-02-19 at 12 17 50

langy:
Resolves #54


Stanton:
:+1:

PR #63, opened 2016-02-23
Change all progress list '-of-type' selectors to '-child' ones instead.

### LINKED ISSUE #62: Progress lists rely on all 'stages' being the same element
Styling is written using `first-of-type` and `last-of-type` selectors. This is fine if they're all the same element type, but if stage cannot be returned to (therefore negating the use of an anchor element and instead using a div, for example), the design breaks.

image

langy:
Resolves #62

### DIFF
--- a/stylesheets/_component.progress-lists.scss
+++ b/stylesheets/_component.progress-lists.scss
@@ -14,7 +14,7 @@
     padding: .75em 0;
     text-decoration: none;
 
-    &:not(:last-of-type) {
+    &:not(:last-child) {
         @include css-arrow(23px, right, $progress-bg);
 
         &:before {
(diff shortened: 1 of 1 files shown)


langy:
Updated docs: https://jadu.gitbooks.io/pulsar/content/progress_list.html#mixing-element-types-for-items

Stanton:
:+1: nice catch!

Stanton:
The explanation in the docs doesn't feel entirely necessary to me given there's a complete markup example, we'd expect there to be unwanted layout issues if people stray from the given markup.

Stanton:
In fact, having a second example which has some items as links and some as divs (or whatever) might be more useful for someone implementing a progress list?

PR #65, opened 2016-02-23
Add disabled option to table rows.

### LINKED ISSUE #64: .is-disabled class needs creating for table rows
Sometimes data needs displaying in a table, but be styled as though it's disabled.

image

langy:
Resolves #64


Stanton:
:+1:

PR #69, opened 2016-02-24
Correctly use `data-toggle="link"` attribute on tabs list if `href` a…

Stanton:
…ttribute is supplied instead of `src`

This makes the helper function as it was originally documented

Also rename the surrounding tests to make their function a bit clearer.


jamesjacobs:
:+1:

Issue #90, opened 2016-03-04
MasterSwitches should use new Toggle styles


Stanton:
- [x] Remove old `toggles` dependency
- [x] Remove old `toggles` styles
- [x] Update component
- [x] Update tests
- [x] Update documentation examples

PR #92, opened 2016-03-04
Improve js code coverage

Stanton:
MasterSwitchComponent is the only file remaining that needs improving, but will need #90 implementing first, which will mean some refactoring is required for the component and tests.


jamesjacobs:
:+1:

PR #97, opened 2016-03-09
Firefox hack to fix occassional misaligned prepended/appended input text.

### LINKED ISSUE #96: Input group addons sometimes have 1px gap (FF)
In Firefox, input group addons are sometimes misaligned. This is likely due to a pixel rounding error, as we currently get around the 'inline-block whitespace bug' using a negative pixel value.

langy:
Resolves #96

### DIFF
--- a/stylesheets/_component.input-groups.scss
+++ b/stylesheets/_component.input-groups.scss
@@ -87,6 +87,9 @@
 .input-group-addon:first-child {
   border-right: 0;
   margin-right: -4px;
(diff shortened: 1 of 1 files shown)


langy:
Ideally we'd use the  fix for these elements but it doesn't look like an easy thing to implement based on the way the macros currently work.

Stanton:
Shouldn't be a problem for macros to fix this... will take a look

PR #99, opened 2016-03-09
Add new get_constant extension to simplify constant checking within views

Stanton:
Checks whether a constant is defined and returns the value, or null if it doesn't exist.

Addresses #6

### DIFF
--- a/src/Jadu/Pulsar/Twig/Extension/GetConstantExtension.php
+++ b/src/Jadu/Pulsar/Twig/Extension/GetConstantExtension.php
@@ -7,6 +7,12 @@ use Twig_SimpleFunction;
 
 class GetConstantExtension extends Twig_Extension
 {
+
+    /**
+     * Register the `get_constant` function with Twig
+     *
+     * @return array The Twig function
+     */
     public function getFunctions()
     {
         return array(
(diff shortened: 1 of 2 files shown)


tompedals:
Looks good to me. Not sure on the `get_constant` name but I can't think of anything better. I assume we can't override the default `constant` Twig function?

Stanton:
Even if we could I'm not sure I'd want to...

src/Jadu/Pulsar/Twig/Extension/GetConstantExtension.php
```diff
@@ -0,0 +1,41 @@
...
+        return 'get_constant_extension';
+    }
+
+    /**
+     * Get constant value if defined, null if not
+     *
+     * @param  string $name The name of the constant
+     * @return string       The value of the constant, null if not defined
+     */
+    public function getConstant($name = '')
```
mgduk: Might it be good to have a fallback value passed by parameter, instead of always returning null? e.g. if a string is wanted. Just add `, $fallback = null` to the args list and `return $fallback` on line 36

Stanton: Sure, can see where this would come in handy :+1:

PR #102, opened 2016-03-10
Update breadcrumb styles

### LINKED ISSUE #101: Add breadcrumb styles for CMS/XFP

Stanton:
Driven by the need for Continuum CMS and XFP to have breadcrumbs restored, especially in legacy UIs.

The breadcrumb styles being removed in this commit are the Pulsar 3's clone of CMS 1.10 styles and are good to be removed rather than deprecated.

`main-title` position has been tweaked to maintain a neat baseline, and also cope with the absence of `.breadcrumb` if it's not there.

screen shot 2016-03-10 at 15 41 52

Closes #101

### DIFF
--- a/views/pulsar/layouts/base.html.twig
+++ b/views/pulsar/layouts/base.html.twig
@@ -36,7 +36,7 @@
     <link rel="stylesheet" type="text/css" href="{{ libsPath ~ 'font-awesome/css/font-awesome-ie7.min.css' }}" />
     <![endif]-->
 
-    <!-- <link rel="stylesheet" href="http://basehold.it/24"> -->
+    <link rel="stylesheet" href="http://basehold.it/24">
 
     <script
         src="{{ base_path ~ 'dist/js/bundle.js' }}"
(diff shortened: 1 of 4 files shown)


views/pulsar/layouts/base.html.twig
```diff
@@ -36,7 +36,7 @@
     <link rel="stylesheet" type="text/css" href="{{ libsPath ~ 'font-awesome/css/font-awesome-ie7.min.css' }}" />
     <![endif]-->
 
-    <!-- <link rel="stylesheet" href="http://basehold.it/24"> -->
+    <link rel="stylesheet" href="http://basehold.it/24">
```
langy: Should we comment out this bad boy again?

langy:
Apart from the bit I commented on, :+1:

views/pulsar/layouts/base.html.twig
Stanton: Nah, I think I'm going to leave it on for the Lexicon from now on

PR #105, opened 2016-03-10
Use cursor pointer when mousing over toggle-switch

### LINKED ISSUE #103: toggle-switch needs to use `pointer` cursor
Current behaviour uses regular pointer which loses some affordance that the toggle can be clicked.

toggle

- [ ] Update codepen styles in documentation

Stanton:
Adds expected affordance that this element can be clicked on.

cursor

Closes #103


langy:
:+1:

PR #110, opened 2016-03-14
Add new status indicator component, helper and tests

### LINKED ISSUE #109: Create new 'status indicator' component
Not quite a label, not quite a badge, but some form of indicator which can indicate online/offline or active/inactive.

- [x] Helper
- [x] Unit tests
- [x] Documentation
- [x] Lexicon

Stanton:
Closes #109 

Documentation for gitbook:
----

# Status

Displays a small visual indication of state; helpful when used within table
cells or piano list keys to avoid the need for a user to check many items
individually to determine their state.

Status indicators should ideally have some text which indicates their context. For example, they could be sat next to a text label called 'Active', or within a table column entitled 'Online'. Try not to rely on this indicator alone.

<p data-height="200" data-theme-id="21222" data-slug-hash="mPrvqZ" data-default-tab="result" data-user="pulsar" class="codepen">See the Pen <a href="http://codepen.io/pulsar/pen/mPrvqZ/">docs - status indicators</a> by Pulsar (<a href="http://codepen.io/pulsar">@pulsar</a>) on <a href="http://codepen.io">CodePen</a>.</p>
<script async src="//assets.codepen.io/assets/embed/ei.js"></script>

### Accessibility

You may notice 'offline' uses a square instead of a circle, this is to visually distinguish it from the green 'online' status for red/green colourblind users as colour alone isn't a reliable indicator of status. The title attribute is also added which is displayed on hover and for screen readers.

Showing how status labels are viewed for Deuternopia sufferers

_Example of how status indicators are seen by deuternopia sufferers_

## Example usage

```twig
{{ html.status('online') }}
```

The 'state' will be used for both the title attribute, and for the stateful
class name, eg:

```html
<span class="status is-online" title="online"></span>
```

### States

 * online
 * offline
 * active
 * inactive

Some states are simply synonyms of each other, the choice of verb will be
defined by the context in which it is to be used.

## Options

Option | Type   | Description
------ | ------ | --------------------------------------------------------------
class  | string | CSS classes, space separated
id     | string | A unique identifier, if required
data-* | string | Data attributes, eg: `'data-foo': 'bar'`

### DIFF
--- a/stylesheets/_component.status.scss
+++ b/stylesheets/_component.status.scss
@@ -3,7 +3,6 @@
     height: 12px;
     width: 12px;
     position: relative;
-    text-indent: -999px;
     top: 1px;
     vertical-align: baseline;
 }
--- a/stylesheets/_component.status.scss
+++ b/stylesheets/_component.status.scss
@@ -1,0 +1,28 @@
+.status {
+    display: inline-block;
+    height: 12px;
+    width: 12px;
+    position: relative;
+    text-indent: -999px;
+    top: 1px;
+    vertical-align: baseline;
+}
+
+.status {
+    &.is-active,
+    &.is-online {
+        background-color: #297c46;
+        border-radius: 6px;
+    }
+
+    &.is-offline {
+        background-color: #c84d40;
+        border-radius: 2px;
+    }
+
+    &.is-inactive {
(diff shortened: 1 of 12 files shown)


langy:
:+1:

PR #118, opened 2016-03-15
Assign additional classes to dropdown button element

### LINKED ISSUE #116: Dropdowns should apply additional classes to button not container
It's expected that any additional classes being applied are much more likely to be for the button element than its container.

langy:
Resolves #116


langy:
Updated PR, now with passing tests :tada:

Stanton:
:+1:  I like the avoidance of an extra space in the classes, though that seems like something we should let the attribute parser handle in the future

Gitbook updated?

Issue #56, opened 2016-02-19
Flash message links and close icon colours are always white


langy:
These should ideally be coloured the same as other elements of that flash message.
![screen shot 2016-02-19 at 14 51 06](https://cloud.githubusercontent.com/assets/713141/13178744/742a41bc-d718-11e5-90c2-5738aaadd9c7.png)

PR #120, opened 2016-03-15
Use `color: inherit;` for flash message links and close buttons.

langy:


langy:
Resolves #56

Stanton:
:+1: 

Could you add screenshots in future? Helps speed up reviews :)

PR #128, opened 2016-03-21
Notifications component

Stanton:
Docs: https://jadu.gitbooks.io/pulsar/content/notifications.html

Also: 
* Remove old deprecated notifications styles using the same namespace
* Add some new styles for media and utilities
* Add new 'new' state colours to base palette
* Add new 'background' alias to base palette

screen shot 2016-03-20 at 21 52 22

notifications


stylesheets/_component.notifications.scss
```diff
@@ -0,0 +1,99 @@
...
+.notifications__header {
+    border-bottom: 1px solid color(border);
+    float: left;
+    padding: 10px;
+    width: 100%;
+}
+
+.notifications-title {
+    display: inline-block;
+    line-height: 1.2em;
```
langy: Do you need the `em` for these line-heights or could you use a unitless ones instead?

langy:
Quite a few instances of hard-coded font sizes and whitespace rather than using vars but assuming this is intentional, so :+1:

Stanton:
Yeah, we need to get some decent vertical rhythm vars into Pulsar to start conforming to the baseline grid

PR #135, opened 2016-03-30
Minor UI tweaks for friendly URLs UI

Stanton:


stylesheets/_component.panels.scss
```diff
@@ -15,6 +15,8 @@
 
     &__body {
         margin-left: 38px;
+        max-height: 20em;
+        overflow-y: auto;
```
langy: Assuming autoprefixer will add --ms-overflow-y here for IE8 compat

langy:
:+1:

stylesheets/_component.panels.scss
Stanton: I though that wasn't required for standards mode?

langy: More Googling makes it seem that way, I felt sure overflow-x and -y had issues in one browser though, maybe it was Firefox and it's no longer a problem.

Unrelated issue with IE8 and combining overflow auto with max-height or max-width though: http://stackoverflow.com/questions/7707/ie8-overflowauto-with-max-height

PR #149, opened 2016-04-07
Move button alignment fix into correct object/partial.

### LINKED ISSUE #147: Move table button alignment fix into .datatable object
_component.tables.scss is no longer in use so the fix needs moving into _components.datatables.scss instead.

langy:
Resolves #147


Stanton:
:+1:

PR #153, opened 2016-04-08
Add z-index: -1; to prevent table-selection TH from hiding TD border.

### LINKED ISSUE #152: TD with 'table-selection' class hides border directly beneath in Firefox
image

This is due to the `position: relative;` property which I believe forces it to sit above the TD on the Z axis.

langy:
Resolves #152


langy:
With negative z-index:

image

Stanton:
:+1:

PR #157, opened 2016-04-13
Notifications implementation

steadweb:


stylesheets/_component.notifications.scss
```diff
@@ -55,6 +34,54 @@
...
+        right: -2px;
+        top: 3px;
+        height: 10px;
+        width: 10px;
+    }
+}
+
+.notifications-list {
+    float: left;
+    max-height: 75vh;
```
jamesjacobs: IE8 fallback needed?

stylesheets/_component.notifications.scss
```diff
@@ -81,14 +108,38 @@
...
+        }
+
+        i {
+            font-size: 3em;
+        }
+    }
+
+    // prevent large images from cropping outside the container
+    img {
+        max-width: 100% !important;
```
jamesjacobs: No way to use a class here / could we chain to lose `!important` or are you overriding inline style?

Stanton: Contents of notification come through RSS so not able to add classes

PR #159, opened 2016-04-19
Make sure notifications dropdown always sits at least 20px from right edge of screen

Stanton:
When site name is very small it was causing the notifications to sit flush

Before:

screen shot 2016-04-19 at 12 33 08

After:

screen shot 2016-04-19 at 12 34 32


langy:
👍

PR #165, opened 2016-04-21
Add ability for form helpers to accept `error` text and display appropriately

### LINKED ISSUE #164: Add 'errors' parameter to form helpers
* Should accept a string, or an array of strings
* It's presence should automatically add the `has-error` styling to the parent form group
* Should not affect text colour of `help` text

Stanton:
Example usage:

```
    form.text({
        'label': 'foo',
        'error': 'AHMAHGAAAAD!'
    })
```

screen shot 2016-04-21 at 10 44 54

To prevent existing form errors from losing error styling, if a form group has error styling, and only a help block, the help block will still be red.

Closes #164


langy:
👍

PR #186, opened 2016-05-04
Flash messages in IE8 should use RGB

### LINKED ISSUE #184: Flash messages are black in IE8
Needs an IE8 rgb fallback

Stanton:
screen shot 2016-05-04 at 09 38 49

Closes #184


jamesjacobs:
👍

PR #190, opened 2016-05-04
Improve styling of file input state variations

### LINKED ISSUE #182: Improve styling of file input state variations
Coloured boxes look misaligned

screen shot 2016-05-03 at 09 48 46

Stanton:
Add extra padding around element only when it's using a state variation so that the default help blocks don't get pushed down too far.

screen shot 2016-05-04 at 09 52 56

Closes #182

### DIFF
--- a/stylesheets/_component.forms.scss
+++ b/stylesheets/_component.forms.scss
@@ -64,7 +64,7 @@ fieldset {
         padding: 0;
 
         @include respond-min($screen-tablet) {
-            padding: 11px 0 0;
+            padding: 14px 0 0;
         }
 
         @include ie-lte(9) {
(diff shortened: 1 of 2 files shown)


langy:
👍

langy:
screen shot 2016-05-11 at 13 56 53

Vertical padding values increased slightly to help with vertical alignment with other form components (e.g. issues addressed within #197)

Stanton:
The disabled example still looks a bit out of line

langy:
It will do unfortunately due to the naked-ness of the input. If it had a bg we could push it down. Otherwise it'll either look too close to, or too far away from, the input. I think being half-a-line is fine in this instance.

Stanton:
Fair point!

PR #194, opened 2016-05-04
Choice block inputs with `selected` should have `is-selected` class applied on page load

### LINKED ISSUE #192: Form choice block not setting correct styling when `checked:true` is set in the options
Using 

    {{
        form.choice({
            'label': '',
            'class': 'choice--block choice--block-inline form__group--medium',
            'optimize': 'few',
            'options': [
                {"label":"Bold","value":"1","checked":true},
                {"label":"Italic","value":"2","checked":true},
                {"label":"Underline","value":"3","checked":true},
                {"label":"Strikethrough","value":"4","checked":true},
                {"label":"Format","value":"5","checked":true}
            ]
        })
    }}

the correct checked styling isn't being applied.

screen shot 2016-05-04 at 15 14 24

Stanton:
Observe new behaviour when page refreshed:

choice-refresh

Closes #192

### DIFF
--- a/js/PulsarFormComponent.js
+++ b/js/PulsarFormComponent.js
@@ -35,7 +35,22 @@ PulsarFormComponent.prototype.init = function () {
 
     // Block styled checkboxes and radios
     var choiceBlock = component.$html.find(".choice--block");
+
+    // set up choice block states on load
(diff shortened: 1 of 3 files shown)


langy:
👍

PR #198, opened 2016-05-11
Increase vertical whitespace to align help block text.

### LINKED ISSUE #197: Form help text doesn't line up with long form label text
When a form label is long enough to span onto a third line, it is noticeable that the form help text doesn't line up with it.

Assuming the form label has the correct line height applied to sit properly within the vertical grid, the solution should be to slightly increase the top margin of the form help text.

langy:
Resolves #197

### DIFF
--- a/stylesheets/_structure.forms.scss
+++ b/stylesheets/_structure.forms.scss
@@ -26,7 +26,7 @@ $control-column-width: 75%;
         margin: 0;
 
         &:nth-of-type(1) {
-            margin-top: .5em;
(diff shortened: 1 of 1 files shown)


langy:
screen shot 2016-05-11 at 13 17 40

Stanton:
File inputs in error state have slightly different placement for help text depending on whether they're using a state modifier, you might need to double check this works with your new alignment...

screen shot 2016-05-04 at 09 52 56

Stanton:
👍

PR #199, opened 2016-05-13
Add js implementations of Pulsar's extensions that test against PHP tests

dave-irvine:


Stanton:
The version of `PulsarFormComponentTest.js` moved to `/web` doesn't match the removed version which has more test cases

Stanton:
👍

PR #200, opened 2016-05-13
Make primary navigation scrollable when it exceeds the viewport height.

### LINKED ISSUE #7: Primary navigation doesn't show all menu items when browser window reduced
In this example, the 'Paybridge' module link is completely inaccessible and cannot be scrolled to when the browser height is reduced so it appears off screen.

screen_shot_2016-01-27_at_14 45 30_720screen_shot_2016-01-27_at_14 45 23_720

langy:
Resolves #7


langy:
scroll-primary-nav

Stanton:
👍 

We'll be able to sort this a nicer way later, maybe collapsing hidden items within a 'more' item or something?

PR #201, opened 2016-05-13
Allow more aspects of the base template to be overridden

dave-irvine:


views/pulsar/layouts/base.html.twig
```diff
@@ -63,45 +63,48 @@
...
+                {% block search %}
                 <input type="search" name="q" id="case-search" value="" placeholder="Search" class="main-search__field" autocomplete="off">
                 <button class="main-search__submit" type="submit" name="case-search"><i class="icon-search"></i></button>
+                {% endblock %}
             </form>
 
             <div class="content-main">
                 <div class="tabs__content">
                     <nav class="breadcrumb">
+                    {% block breadcrumb %}
```
Stanton: Will improve this shortly with #203

views/pulsar/layouts/base.html.twig
```diff
@@ -63,45 +63,48 @@
...
+                {% endblock %}
                 </div>
 
-                {% include '@pulsar/lexicon/includes/notifications.html.twig' %}
+                {% block notifications %}{% endblock %}
 
             </div>
 
             <form action="/search" class="main-search" data-ui="search-box">
+                {% block search %}
```
Stanton: Is the `/search` form action always appropriate, should we include `<form>` within the block too?

dave-irvine: It becomes tricky for consistency :| All the other blocks sit inside the wrapper element, but in this case you're right, it needs to be outside. Unless we add a `div` wrapper that we can put the block inside?

Stanton: It's not a structural element so I'd say putting the entire `<form>` within the block should be ok in this case

PR #205, opened 2016-05-16
Add datatables-select plugin

Stanton:
Add select plugin to Pulsar's browserify bundle and update datatables markup example


dave-irvine:
LGTM

Issue #216, opened 2016-05-26
Add larger size variation of labels


Stanton:
There's a risk of these being too visually similar to buttons at larger sizes, perhaps border-radius should be increased?

<img width="478" alt="screen shot 2016-05-26 at 20 49 07" src="https://cloud.githubusercontent.com/assets/18653/15587936/78f08c58-2383-11e6-93a0-97aed86be503.png">

PR #220, opened 2016-05-27
Add `removable` option to `html.label` helper

Stanton:
- [x] Add option to `html.label` helper
- [x] Icon should inherit label's choice of text colour to maintain AA

```
{{
    html.label({
        label: 'default',
        removable: true
   })
}}
```

screen shot 2016-05-27 at 10 44 51

Normal remove buttons have a `target` attribute, for these we will assume—at this point—that whatever JS is wired up by the developer will simply reference the element being clicked. (the helper still outputs `data-action-target="this"`)

### DIFF
--- a/stylesheets/_component.labels.scss
+++ b/stylesheets/_component.labels.scss
@@ -1,72 +1,84 @@
 .label {
-  @include inline-block;
-  @include content-box;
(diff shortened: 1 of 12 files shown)


Stanton:
Added large variation to this branch. Closes #216

PR #231, opened 2016-06-09
Select2 dropdown should no longer be hidden when select2 is used in a modal

### LINKED ISSUE #230: Select2 dropdown is not visible when used in a modal

Stanton:
… modal

Fixes #230


langy:
👍

PR #237, opened 2016-06-20
Use consistent horizontal spacing between inline navigation tabs

### LINKED ISSUE #236: Improve horizontal alignment of inline navigation tabs with rest of content
At present each tab has `15px` of left and right padding, with `20px` of right margin to separate them.

My proposal is to set each tab to have `20px` of left and right padding, with `10px` of right margin instead.

This will then allow the left edge of every furthest-left textual element on the page to nicely aligned.

Before:
screen shot 2016-06-20 at 17 29 13

After:
screen shot 2016-06-20 at 17 29 36

langy:
Resolves #236


Stanton:
👍

PR #239, opened 2016-06-29
Update no-wrap table class to support new .u- utility format

Stanton:
Keep supporting `.no-wrap`, no plans to deprecate that just yet


jamesjacobs:
Small thing - would recommend adding a silent class/placeholder for `u-no-wrap` and extending that.

Stanton:
@jamesjacobs Added silent placeholder

PR #242, opened 2016-07-01
Add a .gitattributes file

### LINKED ISSUE #238: Add .gitattributes file to reduce size when cloned as a dependency
We include a lot of libs that include tonnes of files that never get used outside of development. These should therefore probably be excluded from the repo when it's cloned as a dependency (via composer, for example).

More info: https://www.reddit.com/r/PHP/comments/2jzp6k/i_dont_need_your_tests_in_my_production

Stanton:
Closes #238


langy:
👍 , however, the `/libs` dir could probably do with being gone through at some point too (it's currently 22mb in size, likely due to loads of files we'll never use in production)

Stanton:
screen shot 2016-07-01 at 11 25 17

langy:
Also may need testing to make sure the directories specified are actually excluded as I've just been reading about potentially requiring a trailing slash; http://stackoverflow.com/questions/5524731/git-ignoring-gitattributes-pattern

langy:
@Stanton Could you sanity check my commit plz?

Stanton:
Looks ok at a glance, I haven't looked at every line of course.

Not sure how to invoke this through composer, tried adding `prefer-dist` to CMS's composer.json and reinstalling composer deps, but nothing changed...

Stanton:
👍

PR #243, opened 2016-07-01
Allow integer 0 values as attributes.

tompedals:


Stanton:
@tompedals `AttributeParserExtension.js` also needs patching for Twig.js

L:33 `if ((value && !_.isArray(value)) || value === '0' || value === 0) {`

tompedals:
@Stanton I've updated the JS version now too. Waiting to see if it all goes green...

PR #245, opened 2016-07-04
Add helper and styles for toolbar link to continuum user manuals

### LINKED ISSUE #244: Create new helper for adding a user guide link in toolbar

Stanton:
Initial version of this is explicitly labelled to help users adopt the feature, once we're confident it's being used we may iterate it down to a simple icon.

screen shot 2016-07-04 at 13 36 36

Closes #244


Stanton:
screen shot 2016-07-04 at 15 01 39

langy:
Looks like something funky's happened with the .gitattributes file, maybe reverted back to an older version? Back-slashes to escape spaces on line 70 didn't work so were replaced with ?'s

langy:
Apart than the above, 👍

jamesjacobs:
👍

.gitattributes
```diff
@@ -66,7 +67,7 @@
 ##
 /libs/html5shiv/dist/html5shiv.min.js                           export-ignore
 /libs/html5shiv/Gruntfile.js                                    export-ignore
-/libs/html5shiv/MIT?and?GPL2?licenses.md                        export-ignore
```
Stanton: @langy So the line should be `/libs/html5shiv/MIT?and?GPL2?licenses.md` ?

langy: Yep, see https://github.com/jadu/pulsar/commit/2b54bc8cbfddc699c692e507397c9ad58936beba

Several other changes were made too

PR #253, opened 2016-07-15
Make datatables automatically render 'no results' message

Stanton:
## No results

If your table has no results in the `tbody` then the table will show a generic message. This message can, and should, be customised to suit the content in question via the `data-empty-table` attribute.

```
<table class="table datatable" data-empty-table="There are currently no documents to display">
    ... 
</table>
```

## Disable row selection

Datatables allow selection of rows by default, this behaviour can be disabled through the `data-selection` attribute. This will hide the related elements in the UI.

```
<table class="table datatable" data-selection="false">
   ...   
</table>
```


jamesjacobs:
👍  

Closes #253

PR #256, opened 2016-07-18
Allow inline navigation links to be disabled

### LINKED ISSUE #233: Add 'disabled' option to inline navigation / tabs
Tabs can be made active, or hidden, but not disabled. It would be nice to be able to inform users of all navigation options from the get-go but just disabling those they can't access yet, rather than magically introducing them.

Example:
image

Stanton:
Example:

```
...
{
    "id"    : "buttons",
    "label" : "Buttons",
    "src"   : tab_buttons,
    "disabled": true
}
...
```

screen shot 2016-07-18 at 14 21 23

Closes #233


langy:
👍

Issue #257, opened 2016-07-25
Table no results action button text colour is incorrect


tompedals:
![screen shot 2016-07-25 at 10 03 09](https://cloud.githubusercontent.com/assets/6132043/17095977/1e8ebcd0-524f-11e6-935b-bc5f04a58e01.png)

The button text colour should be white but is being overridden by some generic table link styles.

Stanton:
@tompedals Fixed in develop

Stanton:
Released in `6.7.1`

PR #258, opened 2016-07-25
Make button colours immutable

Stanton:
Button background/label colours are tightly defined, to allow components to work with link colours without affecting buttons, judicious use of `!important` will allow us to make sure button colours are immutable.

Addresses #257


langy:
👍

PR #275, opened 2016-08-04
Use a solid black background for tooltips in IE8

### LINKED ISSUE #274: No tooltip background colour in IE8
virtualbox_ie8 - win7_01_08_2016_16_59_27

Stanton:
Closes #274

### DIFF
--- a/stylesheets/_component.tooltips.scss
+++ b/stylesheets/_component.tooltips.scss
@@ -32,8 +32,8 @@
   color: $tooltip-color;
   text-align: center;
   text-decoration: none;
-  background-color: rgba(0, 0, 0, .9);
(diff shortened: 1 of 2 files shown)


stylesheets/_component.tooltips.scss
```diff
@@ -32,7 +32,8 @@
   color: $tooltip-color;
   text-align: center;
   text-decoration: none;
-  background-color: $tooltip-bg;
+  background-color: rgba(0, 0, 0, .9);
```
langy: Could we use `rgba(color(black), .9)` here instead, which would make it consistent with the rest of the framework that uses an off-black shade for wherever black is needed?

PR #287, opened 2016-08-20
Change sortable action colour to yellow

### LINKED ISSUE #286: Potentially use yellow/cream rather than blue for dragging/previewing highlight colour
When dragging elements on to a 'drop zone', or when previewing a content change in-context, we tend to use a light blue background colour to highlight these areas. It needs trying first to see how it looks, but I feel a light yellow would work better here.

Stanton:
Closes #286 

(poor gif quality)
sorting


langy:
👍

PR #298, opened 2016-08-28
Move js bundle out of <head> in base template

### LINKED ISSUE #277: Move JS files to end of script instead of in <head>
Y'know, as long as nothing major breaks, of course.

Stanton:
This change is mainly a test to prove that there's no adverse effects with moving the main js bundle out of the `<head>`

Closes #277


langy:
👍

PR #299, opened 2016-08-28
Fix issue where guidance popovers were hidden when used in a modal, and didn't scroll correctly

### LINKED ISSUE #262: Guidance popovers hidden underneath modal
screen shot 2016-07-26 at 10 42 04

### LINKED ISSUE #300: Popovers in modals scroll with the underlying content, not the modal
_Should_ be resolvable through the `data-container` attribute, need to think of the best way for the form helper to pass this to the popover rather than the form group element which is the default behaviour.

popovers

Stanton:
Popovers have a lower z-index than modals, I didn't really want to increase it by default as the default popover behaviour (without `clickover`) would remain visible when a modal is opened and quickly unleash the awesome fury of our test engineers. This change should only increase the z-index of any popovers created after a modal is opened.

Popovers are appended before `</body>` so can’t be `.modal .popover` but handily they’re appended after the modal backdrop, so we can target that as a sibling 🙌

screen shot 2016-08-28 at 21 13 26

Closes #262

A new `guidance-container` attribute can be passed to form helpers which will be passed to the guidance popover's `data-container` attribute, allowing a popover to be bound, for example, to a modal container rather than `body`. Closes #300

### DIFF
--- a/stylesheets/_component.modals.scss
+++ b/stylesheets/_component.modals.scss
@@ -196,3 +196,6 @@
     }
 }
 
+.modal__backdrop.in + .popover {
+    z-index: $zindex-modal + 1;
+}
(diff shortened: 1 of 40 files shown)


langy:
👍

PR #309, opened 2016-09-07
Make prepended/appended buttons look more button'y

### LINKED ISSUE #308: Make prepended/appended buttons look more button'y
Currently look like:

screen shot 2016-09-07 at 11 02 59

Will potentially look like:

screen shot 2016-09-07 at 11 02 45

langy:
Resolves #308 

screen shot 2016-09-07 at 11 36 28

### DIFF
--- a/views/pulsar/v2/helpers/form.html.twig
+++ b/views/pulsar/v2/helpers/form.html.twig
@@ -948,11 +948,19 @@ controls.class   | string | A space separated list of class names
             {% set input_group_classes = input_group_classes ~ ' is-removable' %}
         {% endif %}
 
+        {% if (options.parent.prepend_type is defined and options.parent.prepend_type == 'button') %}
+            {% set input_group_classes = input_group_classes ~ ' has-btn-prepended' %}
+        {% endif %}
+
+        {% if (options.parent.append_type is defined and options.parent.append_type == 'button') %}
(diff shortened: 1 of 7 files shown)


views/pulsar/v2/helpers/form.html.twig
```diff
@@ -948,11 +948,19 @@ controls.class   | string | A space separated list of class names
             {% set input_group_classes = input_group_classes ~ ' is-removable' %}
         {% endif %}
 
+        {% if (options.parent.prepend_type is defined and options.parent.prepend_type == 'button') %}
```
Stanton: If `prepend_type` is not passed, will it default to old styling?

langy: It'll just not print out the classes, or the button, I presume (as it uses the same if statement as on line 965)

langy: Ah, I see your point now. Yes, it will.

Stanton: L:965 is an else, so it should pass the button as a regular `input-group-addon`

langy: You're right. No change in functionality here.

jamesjacobs:
👍

Issue #305, opened 2016-09-05
Dropdown menus don't respond well at mobile sizes


Stanton:
Notifications menu is fine, but site switcher and user menu need to follow same pattern

![screen shot 2016-09-05 at 15 50 47](https://cloud.githubusercontent.com/assets/18653/18251378/c2f8fbda-7380-11e6-8e22-957fe9a4050c.png)
![screen shot 2016-09-05 at 15 50 54](https://cloud.githubusercontent.com/assets/18653/18251383/c57812b0-7380-11e6-9af2-aecb27e93b53.png)
![screen shot 2016-09-05 at 15 51 04](https://cloud.githubusercontent.com/assets/18653/18251386/c7ddaf92-7380-11e6-98ee-4b3cce514d27.png)

PR #312, opened 2016-09-07
Make toolbar dropdowns fit screen width at smaller sizes

Stanton:
Dropdowns in the toolbar now follow the same behaviour as the notifications container.

responsive-dropdowns

Addresses #305

### DIFF
--- a/stylesheets/_component.toolbar.scss
+++ b/stylesheets/_component.toolbar.scss
@@ -24,9 +24,9 @@
 
     .dropdown__menu {
         clear: both;
-        left: 10px;
+        left: $padding-small-horizontal;
         position: absolute;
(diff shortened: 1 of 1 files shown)


stylesheets/_component.toolbar.scss
```diff
@@ -13,6 +13,30 @@
...
+        position: static;
+
+        @include respond-min($screen-xsmall) {
+            position: relative;
+        }
+    }
+
+    .dropdown__menu {
+        clear: both;
+        left: 10px;
```
langy: Use horizontal space vars here instead?

langy:
Apart from ☝️ , 👍

PR #319, opened 2016-09-10
Add basic pikaday integration to Pulsar form component

### LINKED ISSUE #302: Pikaday isn't wired up in the lexicon
Focusing in date inputs doesn't trigger date picker

Stanton:
I've added this to the PulsarFormComponent, so this would take effect on all `form.date()` helpers in the wild, or any text fields using `data-datepicker="true"`. I'm a little concerned that CXM may have their own implementation of pikaday which may conflict, although I suspect they aren't using PulsarFormComponent anyway, so it may be moot.

Closes #302 and fixes the Lexicon example

screen shot 2016-09-10 at 21 50 28


langy:
👍

PR #321, opened 2016-09-13
Allow local use of Proxima Nova in Pulsar and the Lexicon

Stanton:
Jadu developers can now use Proxima Nova in the Lexicon. For licensing reasons, non Jadu developers will not be able to do this unless they add their own licensed files.

# Steps

Check out both the `pulsar` and `pulsar-fonts` repos (pulsar fonts available on Jadu's GitLab instance) to the same directory

eg: `localhost/pulsar` `localhost/pulsar-fonts`

In the pulsar repo, run `make`. This will copy the required pulsar fonts files from `pulsar-fonts/src` to `pulsar/fonts` as well as setting the existing empty `pulsar/fonts/_config.fonts.scss` to not track changes in git to precent accidental comitting of this file.

The lexicon should now be displayed with Proxima Nova 🙌

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,8 @@
 /coverage
 /css
 /dist
+/fonts
+/fonts/_config.fonts.scss
 /tests/logs
 /css
 libs/livereload-js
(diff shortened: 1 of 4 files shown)


langy:
👍

PR #332, opened 2016-09-26
Fix button group border colours when state classes are used.

### LINKED ISSUE #316: Button groups borders aren't displayed correctly
screen shot 2016-09-09 at 11 54 46

Also, specificity of border styles looks to be different in chrome and safari dev tools.

langy:
Resolves #316 

screen shot 2016-09-26 at 11 21 35

Bug was introduced with the commit 4faedf2e which made the default borders be of a higher specificity than those with state classes.


Stanton:
👍

PR #342, opened 2016-09-29
Remove pre-commit git hook and grunt tasks

Stanton:
PHPUnit tests take ~1 minute to run and slow down regular development, these should be run by the developer as part of their own testing process before pushing to GitHub.

The pre-commit hook was also added before we had Travis running build jobs.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -503,7 +503,11 @@ module.exports = function(grunt) {
         'email-build'
     ]);
 
-    grunt.registerTask('post-merge', ['sass:dev', 'browserify', 'email-build']);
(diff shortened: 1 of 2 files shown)


Gruntfile.js
```diff
@@ -503,7 +503,11 @@ module.exports = function(grunt) {
         'email-build'
     ]);
 
-    grunt.registerTask('post-merge', ['sass:dev', 'browserify', 'email-build']);
+    grunt.registerTask('post-merge', [
+        'sass:dev',
+        'browserify',
+        'email-build'
+    ]);
```
langy: <3

PR #348, opened 2016-09-30
Make state classes immutable

### LINKED ISSUE #347: `.is-new` styling doesn't always work
Elements that have their style rules nested deeper than one level will override the state styles, needs making immutable.

screen shot 2016-09-30 at 13 20 44

langy:
Resolves #347 

screen shot 2016-09-30 at 13 32 16


Stanton:
@langy This looks like an ideal candidate for `grunt wraith` 🕵️

PR #358, opened 2016-10-10
Changes required for CMS Maintenance UI

Stanton:
* Make `.bottomless` utility class immutable
* Fix disabled styling on naked buttons (spotted with wraith)
* Add `.box--naked` variant to allow a visually unboxed area to sit underneath a boxed area and maintain the same padding

e.g. here, the checkboxes on the white bg are with a `.box--naked` to match alignment with those within the grey box above.

screen shot 2016-10-10 at 10 45 12

### DIFF
--- a/stylesheets/_utility.utilities.scss
+++ b/stylesheets/_utility.utilities.scss
@@ -1,75 +1,75 @@
 .show-baseline {
-  @include baseline-grid;
+    @include baseline-grid;
 }
 
 %hide,
 .hide {
-  position: absolute;
-  width: 1px;
-  height: 1px;
-  padding: 0;
-  margin: -1px;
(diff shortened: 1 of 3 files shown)


stylesheets/_utility.utilities.scss
```diff
@@ -65,7 +65,7 @@ p.success {
 
 // removes bottom margin from an element
 .bottomless {
-  margin-bottom: 0;
+  margin-bottom: 0 !important;
```
langy: Indent 4 spaces

Stanton: Ah! https://github.com/jadu/pulsar/blob/develop/.scss-lint.yml#L93

PR #373, opened 2016-10-18
Restore outline to checkbox and radio inputs in iOS

### LINKED ISSUE #370: Outlines missing from checkbox & radio on iOS
Spotted on iOS10 (iPhone 6s & iPad 2) and iOS5 (iPad 1), so assume it affects all iOS versions/devices.

### Observed behaviour

img_5665
img_5664

### Default iOS (10) visuals

Don't have to meet these 100%, but at the very least Pulsar checkboxes & radios need outlining on iOS.

img_5661
img_5660

Stanton:
Rather than setting a border on all inputs, they're now only set on text, textarea, password and select/select2 leaving the browser to style checkboxes and radios as they see fit.

Closes #370

## iPad (iOS5)
img_5688
img_5692

## iPhone 4s (iOS9)

img_5689
img_5693

## Galaxy s5 (Android Nougat)

img_5697
img_5695

## Windows Phone 7 (for't laugh)

img_5691
img_5694

### DIFF
--- a/stylesheets/_component.forms.scss
+++ b/stylesheets/_component.forms.scss
@@ -22,7 +22,6 @@ fieldset {
 
 .form__control {
     border-radius: $input-border-radius;
-    border: 1px solid $input-border;
     height: $input-height-base;
     line-height: $input-height-base;
     padding: 0 $padding-base-horizontal;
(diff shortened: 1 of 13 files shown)


stylesheets/_component.forms.scss
```diff
@@ -36,23 +35,25 @@ fieldset {
         }
     }
 
+    &[type="password"],
+    &[type="text"] {
```
langy: Could we have classes for these instead, as we've got them for textareas, selects, checkboxes and radios already? Ideally those we already have would become modifiers of form__control instead of their own class, maybe something to iterate in?

langy:
Also, "Windows Phone 7 (for't laugh)" – I laughed. Good one.

PR #376, opened 2016-10-20
Flash message should no longer have 1px offset in some browsers

Stanton:
Fixes this issue:

screen shot 2016-10-20 at 10 24 20

### DIFF
--- a/stylesheets/_component.flash.scss
+++ b/stylesheets/_component.flash.scss
@@ -96,7 +96,7 @@
     clear: both;
     top: -5px;
     left: 0;
-    margin-bottom: 0;
+    margin: 0 -1px;
     position: relative;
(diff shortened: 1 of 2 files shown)


Stanton:
I've spent an hour and can't figure it out, assumed it was a whitespace bug but can't track it down

langy:
http://stackoverflow.com/questions/31719624/displaytable-div-with-percentage-width-1px-bug

@Stanton ☝️  this is the issue.

Changing `.nav-main` to be 74px rather than 75px might be a better solution, as it masks the problem only when it occurs, instead of misaligning any flash messages that aren't affected.

Stanton:
Cheers @langy, have implemented your change and works fine

PR #378, opened 2016-10-21
Add Wraith dependencies to makefile and tidy things up somewhat

Stanton:
* Make no longer checks for existing versions, we trust in the package managers to deal with that for us.
* Removed xcode command line tools, as make doesn’t even exist without these (and OSX promts for their installation as soon as you run make anyway)
* Added vagrant installation
* Added virtualbox installation
* Fixed path to homebrew installation script, stopped it prompting for y/n confirmation
* Stop brew installs breaking when the package already exists
* Added ImageMagick installation
* Added Wraith installation

Also created a wiki page to track known make related issues: https://github.com/jadu/pulsar/wiki/Make


Stanton:
Bit the bullet and ran `make` on my production machine, only one vagrant-related thing thing needed sorting (added to wiki)

PR #393, opened 2016-10-24
Primary navigation now closes when clicked anywhere outside the element

### LINKED ISSUE #384: Main Menu hide enhancement
When the main menu is expanded and i click outside of the menu container the expanded menu should hide . 

Currently the only way to close the expanded menu is by clicking the small icon "<-" in top expanded menu

Stanton:
Closes #384, requested by @skmetaly 

nav-close

Need to remove the incorrectly named `navMainComponentTest` as part of this ticket and re-add it as a separate release. Attempted simply renaming the file in the past but doesn't seem to stick.


Stanton:
JS tests need to cover new methods, but `navMainComponentTest.js` isn't being picked up by `npm test` due to initial lowercase `n` in the name, This *has* been sorted in the past but has broken again, will investigate tomorrow.

PR #401, opened 2016-10-27
Make badge text colours immutable

### LINKED ISSUE #381: Badge text colour should be immutable
For example, when a badge is in a disabled table row, the text is overridden to grey resulting in the label not meeting AA and sometimes not being visible.

screen shot 2016-10-23 at 11 25 11

related to #379

Stanton:
Tested with:
- Regular badges
- Badges within a `.muted` element
- Badges within an element with a specific text colour
- Badges within buttons
- Badges within disabled buttons

screen shot 2016-10-27 at 10 39 21

Closes #381


stylesheets/_component.badges.scss
```diff
@@ -3,165 +3,90 @@
...
-  &:focus {
-    color: $badge-link-hover-color;
-    text-decoration: none;
-    cursor: pointer;
-  }
+    &:hover,
+    &:focus {
+        color: $badge-link-hover-color;
+        text-decoration: none;
+        cursor: pointer;
```
jamesjacobs: Alpha order rules

PR #402, opened 2016-11-02
Remove the black space at the bottom of the sign in page

tompedals:
screen shot 2016-11-02 at 09 50 20

### DIFF
--- a/stylesheets/_structure.scaffolding.scss
+++ b/stylesheets/_structure.scaffolding.scss
@@ -7,11 +7,6 @@ body {
   height: auto;
 }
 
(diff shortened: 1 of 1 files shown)


stylesheets/_structure.scaffolding.scss
```diff
@@ -30,7 +25,8 @@ body {
   clear: both;
   display: table;
   height: 100%;
-  margin: 0;
+  // Allow any content to clear the fixed footer
+  margin: 0 0 40px;
```
tompedals: Moved the padding to be a margin on the `.container` class so it's not added to the sign in page.

PR #405, opened 2016-11-08
Increase contrast of greeter button vs greeter background

Stanton:
Button used same background colour as the main greeter for some reason, darkened it a couple of % to make it contrast with the bg.

Used in Continuum CMS Blog UI.


jamesjacobs:
A before and after shot would be handy here but looks ok to me.

PR #412, opened 2016-11-15
Table rows can now be sorted if the table has the is-sortable class

Stanton:
Just needs the `.is-sortable` class adding to a table, probably not advisable to use this in conjunction with datatables.

Will be used in Continuum CMS to add sortability to things like download files lists.

Had to qualify the is sortable class into `div.is-sortable` and `.table.is-sortable` so as not to break backwards compatibility with existing sortable items, happy to discuss better ways of doing this.

```
<table class="table is-sortable">
```

sortable-tables

### DIFF
--- a/js/main.js
+++ b/js/main.js
@@ -54,9 +54,25 @@
             }
         });
 
-        $('.is-sortable').sortable({
+        $('div.is-sortable').sortable({
              placeholder: "form__group is-sorting",
              helper: "clone",
+             opacity: 0.9,
+             start: function(e, ui) {
+                $(ui.helper).addClass('is-dragging');
+            }
+        }).disableSelection();
+
+        var fixHelper = function(e, ui) {
+            ui.children().each(function() {
+                $(this).width($(this).width());
+            });
+            return ui;
+        };
+
+        $('.table.is-sortable tbody').sortable({
+             placeholder: "is-sorting",
+             helper: fixHelper,
              opacity: 0.9,
              start: function(e, ui) {
                 $(ui.helper).addClass('is-dragging');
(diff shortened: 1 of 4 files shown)


js/main.js
```diff
@@ -54,7 +54,7 @@
             }
         });
 
-        $('.is-sortable').sortable({
+        $('div.is-sortable').sortable({
```
langy: Shouldn't this be `.form__group` rather than `div`?

stylesheets/_mixin.mixins-to-organise.scss
```diff
@@ -234,3 +234,12 @@ $is-ie: false !default;
         }
     }
 }
+
+@mixin cursor($cursor) {
+    @if $cursor == 'grab' {
+        cursor: move; /* fallback if grab cursor is unsupported */
+        cursor: grab;
+        cursor: -moz-grab;
+        cursor: -webkit-grab;
```
langy: Put prefixed properties before actual property

js/main.js
langy: Guessing not based on the CSS further down, so maybe remove the `form__group` class from the placeholder?

views/playground/sortable/tab.html.twig
```diff
@@ -44,5 +43,78 @@
...
+                document.docx
+            </td>
+            <td class="centered">
+                {{ form.checkbox({'bare': 'true'}) }}
+            </td>
+        </tr>
+
+        <tr>
+            <td>
+                Innocent screensaver
```
langy: Amazing

js/main.js
Stanton: It's not production code, placeholder is required for correct functionality

PR #413, opened 2016-11-17
Prevent prototype library from removing toolbar dropdown elements whe…

Stanton:
…n escape key is pressed

This is down to a conflict with prototype’s `hide()` method which is still present on legacy Continuum UIs.

### DIFF
--- a/js/libs/dropdown.js
+++ b/js/libs/dropdown.js
@@ -98,7 +98,7 @@ var $ = require('jquery');
       if (!$parent.hasClass('open')) return
       $parent.trigger(e = $.Event('hide.bs.dropdown'))
       if (e.isDefaultPrevented()) return
-      $parent.removeClass('open').trigger('hidden.bs.dropdown')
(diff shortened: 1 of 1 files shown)


tompedals:
I think we had this issue in XFP when working on some legacy views. This is the hacky fix we came up with:

```
(function ($) {
    $(function () {
        $('.dropdown').on('hidden.bs.dropdown', function() {
            $(this).show();
        });

        $('[data-toggle="dropdown"]').on('hidden.bs.dropdown', function() {
            $(this).show();
        });

        $('[data-toggle="modal"]').on('hidden.bs.modal', function() {
            $(this).show();
        });

        $('[data-toggle="popover"]').on('hidden.bs.popover', function() {
            $(this).show();
        });

        $('[data-toggle="tooltips"]').on('hidden.bs.tooltip', function() {
            $(this).show();
        });
    });
})(jQuery);
```

Stanton:
Cheers, looks to solve the same issue with prototype adding display:none as an online style 👍

PR #425, opened 2016-11-23
PulsarSortableComponent

### LINKED ISSUE #579: Add keyboard support to sortable behaviour
Currently, the behaviour that is bound to `is-sortable` uses jQueryUI's sortable, and we need to add the ability to reorder rows only using the keyboard (that should both be available by no-js as well as with js)

While we're in there, worth checking screenreader support and if there's anything we can do with aria announcements.

Stanton:
When tabbing into a table the user will now be presented with up/down links which allow them to move rows within a sortable table.

sortablez 08 56 07

## Developer responsibilities

* This relies on table rows having the up/down icons in the markup so that they still work in a non-js context
* Devs will also need to wire up the actual non-js actions to the regular `href` attribute of the up/down links

Closes #579


js/PulsarSortableComponent.js
```diff
@@ -0,0 +1,154 @@
...
+
+var $ = require('jquery');
+
+function PulsarSortableComponent(html, window) {
+    this.$html = html;
+    this.window = window;
+}
+
+PulsarSortableComponent.prototype.init = function () {
+
```
mikedevelops: Can we make this consistent with the other function blocks? (I won't comment on all other occurrences, but there a couple more below) I vote for no line breaks on the first and last line of a function as per line 7.

js/PulsarSortableComponent.js
```diff
@@ -0,0 +1,154 @@
...
+        start: function(e, ui) {
+            $(ui.helper).addClass('is-dragging');
+        }
+    }).disableSelection();
+
+     // Trigger update() on sortupdate event
+     /* istanbul ignore next: difficult to test jQueryUI sortable behaviour */
+    this.$html.find('.table.is-sortable tbody').on('sortupdate', function (e, ui) {
+        var $sortableElement = $(this);
+        update(e, ui, $sortableElement);
```
mikedevelops: Could we get a empty line after the `var` declaration here.

js/PulsarSortableComponent.js
```diff
@@ -0,0 +1,154 @@
...
+    return ui;
+};
+
+PulsarSortableComponent.prototype.addOrder = function() {
+
+    var component = this;
+
+    component.$html.find('.table.is-sortable tr > td:first-of-type').each(function(i) {
+        var $this = $(this),
+            label = $(this).text();
```
mikedevelops: We should use the reference here as we've gone to the trouble of creating it.

`label = $this.text();`

js/PulsarSortableComponent.js
```diff
@@ -0,0 +1,154 @@
...
+
+        i++;
```
mikedevelops: I see what you've done here, I think we should `i + 1` instead when it is used...

`'<span class="sortable__count js-sortable-count">' + (i + 1) + '</span> '`

or for super readability...

```
var count= i + 1;

'<span class="sortable__count js-sortable-count">' + count + '</span> '
```

We're mutating the index by calling `i++` so should this loop ever grow, it could catch someone out quite easily.

js/PulsarSortableComponent.js
```diff
@@ -0,0 +1,154 @@
...
+    });
+};
+
+PulsarSortableComponent.prototype.updateOrder = function() {
+
+    var component = this;
+
+    component.$html.find('.table.is-sortable .js-sortable-count').each(function(i) {
+
+        i++;
```
mikedevelops: as above.

Stanton:
@mikedevelops Changes implemented 👯

PR #444, opened 2016-12-19
Fix issue stopping mochify tests running

Stanton:
Thanks @asmblah!


.travis.yml
```diff
@@ -20,6 +20,8 @@ install:
 
 script:
   - phpunit
+  # npm run coverage
```
asmblah: @Stanton Can this coverage step be re-enabled now, or is it handled elsewhere?

PR #455, opened 2017-01-09
CXM - sign in branding and responsive changes to slider component

jamesjacobs:


stylesheets/_component.slider.scss
```diff
@@ -1,45 +1,65 @@
 .slider-container {
     margin: 5em auto 0;
-    width: 600px;
+    padding: 1em;
```
Stanton: Not better with a normal padding variable?

Issue #469, opened 2017-01-23
Add Markup Validators in Grunt


nvasileiadis:
Check all the output files for errors and fix them.

nvasileiadis:
This is the plugin used "https://www.npmjs.com/package/grunt-w3c-html-validation"

PR #470, opened 2017-01-24
Grunt W3C Validation

nvasileiadis:
The validation of the project's output markup HTML with W3C standards validator has been added to and updated the Grunt workflow.


nvasileiadis:
There is a need for a directories check and addition if anything is missing.

Gruntfile.js
```diff
@@ -474,6 +474,25 @@ module.exports = function(grunt) {
...
+                maxTry: 3,
+                relaxerror: ['Bad value X-UA-Compatible for attribute http-equiv on element meta.'], //ignores these errors 
+                generateReport: true,
+                errorHTMLRootDir: "w3cErrorFolder",
+                useTimeStamp: true,
+                errorTemplate: "w3cErrorFolder/w3c_validation_error_Template.html"
+            },
+            files: {
+                src: ['dist/views/*.html', 
+                                'emails/dist/*.html',
```
Stanton: Tabbing looks a bit odd here

Gruntfile.js
```diff
@@ -474,6 +474,25 @@ module.exports = function(grunt) {
...
                 watchTask: true
             }
+        },
+
+        validation: {
+            options: {
+                reset: grunt.option('reset') || false,
+                stoponerror: false,
+                maxTry: 3,
+                relaxerror: ['Bad value X-UA-Compatible for attribute http-equiv on element meta.'], //ignores these errors
```
Stanton: Minor point, but need space after `//`

.gitignore
```diff
@@ -26,3 +26,5 @@ pulsar.zip
 /fonts/_config.fonts.scss
 
 /.csscomb.json
+/validation-status.json
```
Stanton: These files will also need adding to the `.gitattributes` file as `export-ignore` rules so that Continuum products can ignore them

Gruntfile.js
```diff
@@ -474,6 +474,25 @@ module.exports = function(grunt) {
...
+        validation: {
+            options: {
+                reset: grunt.option('reset') || false,
+                stoponerror: false,
+                maxTry: 3,
+                relaxerror: ['Bad value X-UA-Compatible for attribute http-equiv on element meta.'], //ignores these errors 
+                generateReport: true,
+                errorHTMLRootDir: "w3cErrorFolder",
+                useTimeStamp: true,
+                errorTemplate: "w3cErrorFolder/w3c_validation_error_Template.html"
```
Stanton: This might be more appropriate to put in the `/tests` directory, `/tests/validation` perhaps?

Gruntfile.js
```diff
@@ -474,6 +474,25 @@ module.exports = function(grunt) {
...
+        },
+
+        validation: {
+            options: {
+                reset: grunt.option('reset') || false,
+                stoponerror: false,
+                maxTry: 3,
+                relaxerror: ['Bad value X-UA-Compatible for attribute http-equiv on element meta.'], //ignores these errors 
+                generateReport: true,
+                errorHTMLRootDir: "w3cErrorFolder",
```
Stanton: Wherever the output is placed, we'll need to add it to the `.gitignore` file so it's not committed to the repository

Gruntfile.js
```diff
@@ -494,16 +513,19 @@ module.exports = function(grunt) {
         ]
     });
 
+    grunt.loadNpmTasks('grunt-w3c-html-validation');
```
Stanton: Shouldn't need this, it's handled by `matchdep` at the bottom of the gruntfile

Gruntfile.js
```diff
@@ -494,16 +513,19 @@ module.exports = function(grunt) {
...
+
     grunt.registerTask('default', [
```
Stanton: How long does the validation task take? Is it something we want to run on the default task, or something to roll into a testing/pre-release task?

Gruntfile.js
nvasileiadis: Fixed

Gruntfile.js
nvasileiadis: Added

Gruntfile.js
nvasileiadis: Relocated as advised

Gruntfile.js
nvasileiadis: Fixed

Gruntfile.js
nvasileiadis: Deleted

Gruntfile.js
nvasileiadis: It takes some time, changed to 'grunt validate'

.gitignore
nvasileiadis: Added

Gruntfile.js
```diff
@@ -474,6 +474,25 @@ module.exports = function(grunt) {
...
+                maxTry: 3,
+                relaxerror: ['Bad value X-UA-Compatible for attribute http-equiv on element meta.'], // ignores these errors 
+                generateReport: true,
+                errorHTMLRootDir: "tests/validation",
+                useTimeStamp: true,
+                errorTemplate: "tests/validation/w3c_validation_error_Template.html"
+            },
+            files: {
+                src: ['dist/views/*.html', 
+                    'emails/dist/*.html',
```
Stanton: On second thoughts I don't think validating emails is much use as most of this is automatically generated based on email client quirks.

Gruntfile.js
```diff
@@ -474,6 +474,25 @@ module.exports = function(grunt) {
...
+                stoponerror: false,
+                maxTry: 3,
+                relaxerror: ['Bad value X-UA-Compatible for attribute http-equiv on element meta.'], // ignores these errors 
+                generateReport: true,
+                errorHTMLRootDir: "tests/validation",
+                useTimeStamp: true,
+                errorTemplate: "tests/validation/w3c_validation_error_Template.html"
+            },
+            files: {
+                src: ['dist/views/*.html',
```
Stanton: I think we should list the relevant lexicon urls we'd want to validate rather than just the `.html` files e have in `/dist`, this will let us validate against our Twig output.

nvasileiadis: Do you mean /dist/views/pulsar? Because all the other files I see are assets. If this is the only folder with the html and twig parts I will change and commit it.

Gruntfile.js
nvasileiadis: Emails have been removed from the related settings in gruntfile.

js/casper.js
```diff
@@ -0,0 +1,54 @@
...
+    // Remove links within the same page
+    for (var i = 0; i < links.length; i++) {
+        if (links[i].indexOf('#')  == -1) {
+            newlinks.push(links[i]);
+        };
+    };
+
+    // Get Page Name from URL
+    for (var i = 0; i < newlinks.length; i++) {
+        if (newlinks[i].match(regex) != null ) { // Remove null items that didn't pass the regex
```
mikedevelops: This could be simplified further to `if (newlinks[i].match(regex ))` as we'll get an array with the matching regex query if we've got a match, this will also keep the hound at bay.

js/casper.js
```diff
@@ -0,0 +1,54 @@
...
+    var current = 0;
+    var end = links.length;
+    // Get the word before .html.twig, the word before the last '/' and the word in urls like '/lexicon'
+    var regex = /((\w*)(\/|.html.twig)$)|(\/\w*$)/gm;
+    var match = [];
+    var newlinks = [];
+
+    // Remove links within the same page
+    for (var i = 0; i < links.length; i++) {
+        if (links[i].indexOf('#')  == -1) {
```
mikedevelops: There's also an extra space before `==`, I agree with the hound on this one, always best to be explicit with operators.

nvasileiadis: An extra '=' was added

js/casper.js
nvasileiadis: Refactored

nvasileiadis:
Closes #469

nvasileiadis:
Closes #469

PR #473, opened 2017-01-24
Remove immutability of base label variant

### LINKED ISSUE #472: Base variations of labels and buttons need not be immutable
Continuum CXM needs the ability to override the base version of buttons/labels with customisable colour schemes, because all variations currently use `!important` this is not easy without over-specifying the custom styles.

Consider dropping the immutability of the base variant.

Stanton:
This change stops the default `.label` styling from using `!important`, allowing background and text colours to be specified through alternative or inline styles.

Closes #472


nvasileiadis:
After reading the changes here, searching for label related declarations and !important in the sass files. I believe this issue is ready for merge.

PR #485, opened 2017-02-16
CXM - full screen modals

jamesjacobs:
CXM has a need for full page modals, typically for timeline image previews.

To avoid rewriting the existing modal JS (although it really does need some love) I've added a modifier and some additional mark up to create the effect based on the standard modal behaviour.

Also fixed an issue where you could scroll body when viewing a standard modal, which had an odd effect.

Browser tested in Chrome, Safari Firefox, Edge - IE9.

modals


stylesheets/_component.modals.scss
```diff
@@ -3,159 +3,161 @@
 // --------------------------------------------------
 
 // .modal--open      - body class for killing the scroll
-// .modal           - container to scroll within
+// .modal            - container to scroll within
 // .modal__dialog    - positioning shell for the actual modal
 // .modal__content   - actual modal w/ bg and corners
+// .modal--fs        - modifier for a full screen modal
```
Stanton: Is this an alias? isn't your class for full screen modals `.modal--full-screen`?

jamesjacobs: Ah, yes - I change the class name. Will amend.

PR #493, opened 2017-03-01
Remove extra padding when using a control label within a help block

Stanton:
Improves the appearance of bare `checkbox_inline` elements within the `help-block`

Before:

screen shot 2017-03-01 at 11 11 22

After:

screen shot 2017-03-01 at 11 10 56

### DIFF
--- a/makefile
+++ b/makefile
@@ -70,6 +70,7 @@ build:
 	@ echo "\n${CHECK} Done"
 
 	@ echo "${HR}\nCopy Proxima Nova (if available)...${HR}"
+	@ mkdir ./fonts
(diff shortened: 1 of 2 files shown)


makefile
```diff
@@ -70,6 +70,7 @@ build:
 	@ echo "\n${CHECK} Done"
 
 	@ echo "${HR}\nCopy Proxima Nova (if available)...${HR}"
+	@ mkdir ./fonts
```
Stanton: Fixes a make issue where an error was thrown if the `/fonts` directory didn't exist, I thought `touch` would create the directory but it doesn't...

PR #502, opened 2017-03-08
Fix issue where `.btn--white` was being compressed to `.btn--#fff`

### LINKED ISSUE #501: Compressed css output breaks btn--white styles
The grunt-sass task is converting `.btn--white` to `.btn--#fff`

**Steps to replicate:**

* Run `grunt sass:dist_modern`
* File `dist/css/pulsar.css` contains references to `.btn--#fff` instead of `.btn--white`

Stanton:
From reading https://github.com/sass/sass/issues/363 this only affects class names when you're declaring them via a mixin, which is why `.swatch--white` was unaffected. Also only affects comressed output, which is why it wasn't caught in develop.

Solution involves setting a variable to a string and then interpolating it into the classname.

Closes #501


stylesheets/_component.buttons.scss
```diff
@@ -33,6 +33,7 @@ $btn-padding-x:             14px;
 $btn-small-padding-top:     $btn-padding-top / 2.5;
 $btn-small-padding-bottom:  $btn-padding-bottom / 2.5;
 $btn-small-padding-x:       $btn-padding-x / 2.5;
+$btn-white:                 "white";
```
Stanton: Can't, has to be double, will switch off linter for this rule...

PR #503, opened 2017-03-09
Don't enable elements in the masterswitch that were already disabled within the view

### LINKED ISSUE #477: Not currently possible to keep certain elements disabled after masterswitch enabled
The current behaviour of a masterswitch disables all interactive child elements, and enables them again when you enable the masterswitch.

This means it's not possible to keep any elements/sections disabled when the masterswitch enabled.

Perhaps the masterswitch should stop using `is-disabled` and use a different class, allowing the any elements with `is-disabled` to maintain their disabled state.

Stanton:
This allows a view to keep certain element disabled after the master switch has been enabled, such as integrations which need credentials supplying before other fields are enabled.

It's completely backwards compatible and doesn't require any changes to masterswitch components in the wild.

masterswitch

Closes #477

### DIFF
--- a/tests/js/web/MasterSwitchComponentTest.js
+++ b/tests/js/web/MasterSwitchComponentTest.js
@@ -128,7 +128,7 @@ describe('MasterSwitch component', function() {
 
 			setTimeout(function(){
 				expect(clickEvent.isDefaultPrevented()).to.be.false;
-            }, 100);
+            }, 1000);
 		});
 
 		it('should remove the disabled attribute from buttons', function() {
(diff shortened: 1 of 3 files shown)


tests/js/web/MasterSwitchComponentTest.js
```diff
@@ -25,6 +25,7 @@ describe('MasterSwitch component', function() {
 		<a href="#">foo</a>\
 		<button>foo</button>\
 		<input type="text" />\
+		<input type="text" class="is-disabled disabled" />\
```
jamesjacobs: Does this need both classes?

PR #506, opened 2017-03-12
505_select2-x-does-not-render-in-ie11-or-below

jamesjacobs:


jamesjacobs:
I've removed the :after IE was having an issue with and added styles to the X to achieve the same visual appearance.

screen shot 2017-03-12 at 10 04 55

Tested in chrome, safari, firefox and IE edge - IE9.

jamesjacobs:
Note, due to lack of Pulsar fonts in the lexicon they appear slightly broken.
screen shot 2017-03-12 at 10 09 14

Stanton:
Ha, I've just done PR #507 restores the original icons, need to discuss what we think is the best solution, possibly this one...

jamesjacobs:
Decided to go with @Stanton's fix #507

PR #507, opened 2017-03-12
Restore remove icons for Select2 elements in IE

### LINKED ISSUE #505: Select2 X doesn't render in IE11 or below
screen shot 2017-03-10 at 15 29 44

Stanton:
Issue down to the way IE handles visibility on parents. Seems you can't set a parent to be hidden and a child to be visible.

Fixed by 'hiding' the original X character by setting it to be the same colour as the background and stopping it from becoming visible again through hover.

Tested in IE11 and IE9 (My IE10 VM isn't working at the moment)

screen shot 2017-03-12 at 10 03 53
screen shot 2017-03-12 at 10 09 38

Closes #505

### DIFF
--- a/stylesheets/_component.select2.scss
+++ b/stylesheets/_component.select2.scss
@@ -275,21 +275,18 @@
 
 
 .select2-container--default .select2-selection--multiple .select2-selection__choice__remove {
+    color: #dedede;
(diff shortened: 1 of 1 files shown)


jamesjacobs:
Closes #505

PR #509, opened 2017-03-14
Changes required for XFPs Event Rebooking feature

Stanton:
99% of this diff is `/libs`

Main change that is outside of rebooking specific files:

https://github.com/jadu/pulsar/pull/509/commits/84ebd49478e3cd5527d336c786b30fc7d95aaa91 Panels within form groups now have a default width (xlarge) and will respect the width modifier given to the parent form group

screen shot 2017-03-15 at 10 23 13


js/PulsarFormComponent.js
```diff
@@ -5,6 +5,7 @@ var $ = require('jquery');
 require('../libs/pikaday/plugins/pikaday.jquery');
 require('../libs/select2/dist/js/select2.min');
 require('../libs/spectrum/spectrum');
+var moment = require('../libs/moment/moment');
```
jamesjacobs: Needed?

PR #521, opened 2017-03-24
Fix unnecessary scrolling

### LINKED ISSUE #498: Body scrolls when unneeded
Body is scrollable even when it doesn't need to be. In older IE and Older OSX versions this means a vertical scroll bar is always present.

scrolling body no scroll bar due to lastest osx
scrolling body ie9

jamesjacobs:
Fixed by removing footer height from `.inner-wrapper` height.

safari
ie9

Tested in:
- [x] IE9-Edge
- [x] Safari
- [x] Chrome
- [x] Firefox 

Fixes #498

### DIFF
--- a/stylesheets/_layout.tabs.scss
+++ b/stylesheets/_layout.tabs.scss
@@ -29,7 +29,7 @@
     -webkit-transition: -webkit-transform 500ms ease;
     backface-visibility: hidden;
     display: table;
-    height: 100vh;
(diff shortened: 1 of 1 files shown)


Stanton:
Where's the footer in the second screenshot?

jamesjacobs:
Not sure what happened there, my IE VM went a bit crazy when I resized it a few times for the gif. The footer is there though! Have another gif:

ie9

Issue #313, opened 2016-09-08
Decide whether or not various overrides are needed anymore


langy:
There are a number of overrides inside the `_override.shame.scss` file that are quite broad in terms of what elements they're targeting.

This is causing some funky layout issues to occur (see attached screenshot).

The lines needing review are potentially from an old iteration of the Q software, specifically:
https://github.com/jadu/pulsar/blob/develop/stylesheets/_override.shame.scss#L62-L177

![image](https://cloud.githubusercontent.com/assets/713141/18355875/83e070ea-75e2-11e6-843e-706a3c2fa7a6.png)

langy:
![image](https://cloud.githubusercontent.com/assets/713141/18477344/43b8b8a6-79c5-11e6-8260-07418b02b06a.png)

jamesjacobs:
Fixed in https://github.com/jadu/pulsar/pull/522

PR #522, opened 2017-03-27
313_shame-overrides

jamesjacobs:


stylesheets/_component.forms.scss
```diff
@@ -6,6 +6,32 @@ fieldset {
 
 .form {
     background-color: color(white);
+
+    // Added by @stanton, not used in CXM - safe to remove?
+    .slide-inner & {
+        padding-top: 17px;
```
jamesjacobs: Is this still required @Stanton?

Stanton: I'm almost certain `.slide-inner` is a class from the old decks component, I can't find it in use in CMS or Pulsar so should be ok to remove

jamesjacobs:
Fixes #313

jamesjacobs:
Closes #313

PR #524, opened 2017-03-28
Disabled ui component

jamesjacobs:
Allows you to disable parts of a UI by adding the data-disable-ui="true" data attribute.
Form elements are disabled, links can not be clicked and a faded overlay is added.

disabled ui

### DIFF
--- a/js/DisableUiComponent.js
+++ b/js/DisableUiComponent.js
@@ -38,18 +38,7 @@ DisableUiComponent.prototype.init = function () {
             .addClass('u-cursor-not-allowed');
 
         // Wrap with disabled wrapper to visually disable
-        // Check if masterswitch as we don't want to add 2 overlays
-        if ($this.hasClass('masterswitch')) {
-            // If the master switch is enabled, wrap in disabled overlay
-            if ($this.find('.form__control.toggle-switch').is(':checked')) {
-                $this.wrap('<div class="u-ui-disabled"></div>');
-            } else {
-                // If not, only wrap the control and let the master switch add the overlay
-                $this.find('.masterswitch-control').wrap('<div class="u-ui-disabled"></div>');
-            }
-        } else {
-            $this.wrap('<div class="u-ui-disabled"></div>');
-        }
+        $this.wrap('<div class="u-ui-disabled"></div>');
     });
 };
 
--- a/js/DisableUiComponent.js
+++ b/js/DisableUiComponent.js
@@ -4,7 +4,7 @@ var $ = require('jquery');
 
 function DisableUiComponent(html) {
     this.$html = html;
-};
+}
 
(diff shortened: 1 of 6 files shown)


js/DisableUiComponent.js
```diff
@@ -0,0 +1,56 @@
...
+            LABEL_ELEMENTS = 'label';
+
+        // Disable form elements
+        $this.find(FORM_ELEMENTS)
+            .on('click', preventDefaultAndStopPropagation)
+            .addClass('disabled')
+            .attr('disabled', 'disabled');
+
+        // Disable labels
+        $this.find(LABEL_ELEMENTS).addClass('u-cursor-not-allowed');
```
Stanton: is this not something the `.is-disabled` class could handle? We use that quite often and it's a common state variation for a given element, or a group of elements. Same for all instances of `.u-cursor-not-allowed`.

js/DisableUiComponent.js
```diff
@@ -0,0 +1,56 @@
...
+
+        // Disable links
+        $this.find(LINK_ELEMENTS)
+            .on('click', preventDefaultAndStopPropagation)
+            .addClass('u-cursor-not-allowed');
+
+        // Wrap with disabled wrapper to visually disable
+        // Check if masterswitch as we don't want to add 2 overlays
+        if ($this.hasClass('masterswitch')) {
```
Stanton: Any way we can uncouple these? wondering if it could be solved through styling alone and not need to know about the masterswitch?

stylesheets/_utility.utilities.scss
```diff
@@ -301,3 +301,12 @@ div.no-results {
 .u-text-align-justify {
     text-align: justify;
 }
+
+.u-ui-disabled {
+    cursor: not-allowed;
```
Stanton: Could use `@extend %u-cursor-not-allowed;`?

stylesheets/_utility.utilities.scss
```diff
@@ -301,3 +301,12 @@ div.no-results {
...
+    opacity: .25;
+}
+
+.u-cursor-not-allowed {
```
Stanton: Add `%u-cursor-not-allowed` for easier extends

js/DisableUiComponent.js
jamesjacobs: I considered that. The issue being `.is-disabled` has different styles set depending on the element/class it's attached to. For example, some places it reduces color contrast, other places it adds background colors, these things resulted in lots of different opacities/styles being set on the disabled UI contents. I wanted a unified disabled UI with just the cursor and the faded opacity. It doesn't deal with the cursor either, which of course we could add. Moving to a separate specific "one use" class felt better. Thoughts?

PR #535, opened 2017-04-11
QU-1758 drag and drop upload

mikedevelops:
Drag and drop component.

**Vanilla example**
dropzone-vanilla-drop

**Passive example**
dropzone-passive-drop

**Product example**
dropzone-passive-enriched


stylesheets/_component.dropzone.scss
```diff
@@ -0,0 +1,319 @@
...
+    text-align: center;
+}
+
+.dropzone__icon--small {
+    font-size: 1em;
+    position: absolute;
+    will-change: transform;
+}
+
+$dropzone-icon-rotation: 30deg;
```
jamesjacobs: Any reason this isn't at the top with the others?

stylesheets/_component.dropzone.scss
```diff
@@ -0,0 +1,319 @@
...
+.dropzone__label {
+    font-family: $font-family-regular;
+    font-size: 1.25em;
+    line-height: 1;
+    margin-bottom: 0;
+    position: relative;
+}
+
+.dropzone__info {
+    font-size: .85em;
```
jamesjacobs: Could we not use the font size vars here?

stylesheets/_component.dropzone.scss
```diff
@@ -0,0 +1,319 @@
...
+    margin-bottom: 0;
+    position: relative;
+}
+
+.dropzone__help {
+    color: color(text, help);
+    display: inline-block;
+    font-size: .85em;
```
jamesjacobs: and here

stylesheets/_component.dropzone.scss
```diff
@@ -0,0 +1,319 @@
...
+}
+
+/**
+ * Error
+ */
+
+.dropzone__error {
+    color: color(danger);
+    font-family: $font-family-regular;
+    font-size: .85em;
```
jamesjacobs: font size var?

stylesheets/_component.dropzone.scss
```diff
@@ -0,0 +1,319 @@
...
+    line-height: 1.1;
+    margin-bottom: $dropzone-file-size;
+}
+
+/**
+ * State
+ */
+
+@mixin dropzone-flash ($state) {
```
jamesjacobs: could this be more generic for reuse? and moved to the mixin partial?

views/lexicon/forms/dropzone.html.twig
```diff
@@ -0,0 +1,38 @@
...
+        })
+    }}
+
+{% endblock tab_content %}
+
+{% block tab_sidebar %}
+    <a href="https://jadu.gitbooks.io/pulsar/content/text.html">
+        <img src="../../../images/cover.png" alt="Pulsar documentation front cover" width="157px" />
+    </a>
+    <br />
```
jamesjacobs: Could use a margin utility class here rather than a br

stylesheets/_component.dropzone.scss
```diff
@@ -0,0 +1,311 @@
...
+    position: absolute;
+    pointer-events: none;
+    right: 0;
+    top: 0;
+    width: 100%;
+    z-index: 0;
+}
+
+.dropzone__file-wrapper {
+    display: flex;
```
mikedevelops: I should probably write some fallback code in, at the moment it falls back to everything being block, which looks acceptable, but should probably get some more love

jamesjacobs:
Took it for a ride in IE9 to Edge, Chrome, Safari and FF. Found a few issues:

When I drag a file in and out of the window a few times:
dragged in and out of window fast

Missing popover title:
title

Line height is a bit cramped here (on the error text):
screen shot 2017-07-10 at 13 32 27

When I accidentally tried to drag lots of files in, I got this and couldn't add any more:
screen shot 2017-07-10 at 13 31 47

IE9 support - can't remember what was decided on this but I thought we would just show either a standard file input or remove text that mentioned drag and drop:
ie9

Finally, in the lexicon, I'm not sure it belongs in html helpers, feels like it should be in Modules. Also the page title is incorrect.
page title nav and section

views/lexicon/forms/dropzone.html.twig
mikedevelops: This is a fresh copy from a Lexicon tab content, we should probably make this a twig partial and remove the little underline on the image (I was going to do that but forgot!)

stylesheets/_component.dropzone.scss
mikedevelops: I've moved the file size variables a little further down for consistency, I prefer having config variables in throwing distance from their context, rather than the put all variables at the top approach.

stylesheets/_component.dropzone.scss
mikedevelops: I've double checked this and the fallback is all good. IE9 is the only browser that is unsupported - but it also doesn't get the `file-wrapper` by default. In CXM I've mocked proper file objects for consistency, so the fallback code is part of that ticket.

stylesheets/_component.dropzone.scss
mikedevelops: It _could_, I'm not sure it should though. It needs to use the `color(...)` function which is included after we include our base mixins, so it'd throw out the consistency there, I'm also not sure it's generic enough to warrant being an "official" mixin.

js/DropZone/DropZoneValidatorDispatcher.js
```diff
@@ -0,0 +1,88 @@
...
+        let sizeCount = totalSize;
+
+        // if we have files but we do not have a length we're dealing
+        // with a browser with limited support, so we'll return them as valid
+        // now and let them be caught later
+        if (!files.length) {
+            return result;
+        }
+
+        for (let i = 0; i < files.length; i++) {
```
asmblah: Could this be a `for..of`?

js/DropZone/DropZoneValidatorDispatcher.js
```diff
@@ -0,0 +1,88 @@
...
+            if (
+                result.valid &&
+                this.whitelist &&
+                this.whitelist.length &&
+                !file.mock &&
+                !this.utils.validateType(file.type, this.whitelist)
+            ) {
+                result = this.errors.getFileValidationError('WHITELIST');
+            }
+
```
asmblah: Extra newline 🍹

js/index.js
```diff
@@ -43,22 +43,24 @@ var $                     = require('jquery'),
...
-    SignInComponent = require('./area/signin/signin');
+    SignInComponent = require('./area/signin/signin'),
+    DropZoneComponent = require('./DropZone/DropZoneComponentFactory');
 
     require('jstree');
 
 module.exports = {
-    ButtonComponent: ButtonComponent,
+    ButtonComponent,
+    DropZoneComponent,
```
asmblah: Should we stick `Factory` on the end here (and above)?

js/libs/MimeTyper.js
```diff
@@ -0,0 +1,95 @@
...
+    /**
+     * Search MIME keys and return corresponding class name.
+     * If we can't find the type, return the default class name.
+     * @param  {String} userMime
+     * @return {String}
+     */
+    getIconClass (userMime) {
+        let match;
+
+        if (Array.prototype.find) {
```
asmblah: This seems like a bit of a micro-optimisation - should we just use `_.find(...)` until/unless we find this is a performance bottleneck? Just to keep the logic simple(r)

tests/js/web/DropZoneBodyClassManagerTest.js
```diff
@@ -0,0 +1,23 @@
...
+
+    beforeEach(() => {
+        $body = $('<div class="foo"></div>');
+        classManager = new DropZoneBodyClassManager();
+    });
+
+    describe('update()', () => {
+        it('should update the body class with the new state', () => {
+            classManager.update($body[0], ['bar', 'baz']);
+            expect($body.attr('class')).to.equal('foo bar baz');
```
asmblah: Personal preference, but I find it nice to have a blank line between the sections:
```javascript
// Arrange (although I realise there is no "arrange" section is this particular test)
this.thing.configure(...);

// Act
this.thing.doSomething();

// Assert
expect(this.otherThing).to.have.been.calledOnce;
```

tests/js/web/DropZoneCallbackManagerTest.js
```diff
@@ -0,0 +1,27 @@
...
+    beforeEach(() => {
+        callbackManager = new DropZoneCallbackManager();
+        callbackSpy = sinon.spy();
+        dropZoneSpy = sinon.spy();
+    });
+
+    describe('create()', () => {
+        it('should merge the data with a reference to the instance and invoke callback', () => {
+            callbackManager.create(callbackSpy, dropZoneSpy, { foo: 'bar' });
+            expect(callbackSpy).to.have.been.calledWith({ foo: 'bar', instance: dropZoneSpy });
```
asmblah: I think if you want to check that the correct instance is passed in, you want `instance: sinon.match.same(dropZoneSpy)` instead

tests/js/web/DropZoneComponentTest.js
```diff
@@ -0,0 +1,928 @@
...
+            inputStub = sinon.stub(dropZoneComponent, 'processInputNode');
+            browseStub = sinon.stub(dropZoneComponent, 'processBrowseNode');
+            instanceManager.getInstance.returns([
+                { id: 0 },
+                { id: 1, input: {}, options: {}, browse: {} }
+            ]);
+        });
+
+        afterEach(() => {
+            $html.find('.dropzone.test').remove();
```
asmblah: I don't _think_ this is needed, as you create a new `$html` for every test above, which should be creating a new `<html>` element and discarding the old one (and all its descendants).

tests/js/web/DropZoneComponentTest.js
```diff
@@ -0,0 +1,928 @@
...
+
+        // partial stub
+        it('should update info node with the idleHtml option', () => {
+            optionsManager.getInstanceOptions.returns({
+                idleHtml: 'foo',
+                nodeClasses: {}
+            });
+            dropZoneComponent.throwValidationError('foo', 0);
+            expect(partialStub).to.have.been.calledOnce;
+            partialStub.reset();
```
asmblah: No need for this reset, a new stub should be created for each test above

tests/js/web/DropZoneComponentTest.js
```diff
@@ -0,0 +1,928 @@
...
+
+    describe('throwValidationError()', () => {
+        let partialStub;
+
+        beforeEach(() => {
+            partialStub = sinon.stub(dropZoneComponent, 'updateInfoState');
+        });
+
+        afterEach(() => {
+            partialStub.reset();
```
asmblah: No need for this reset, a new `DropZoneComponent` should be created above and the old one should be GC'd.

tests/js/web/DropZoneComponentTest.js
```diff
@@ -0,0 +1,928 @@
...
+        let infoStub;
+        let throwStub;
+
+        beforeEach(() => {
+            infoStub = sinon.stub(dropZoneComponent, 'updateInfoState');
+            throwStub = sinon.stub(dropZoneComponent, 'throwValidationError');
+        });
+
+        afterEach(() => {
+            infoStub.reset();
```
asmblah: Same with these

tests/js/web/DropZoneComponentUtilsTest.js
```diff
@@ -0,0 +1,217 @@
...
+    });
+
+    describe('camelCasify()', () => {
+        it('should transform a hyphen serrated string to camel case', () => {
+            expect(utils.camelCaseIfy('foo-bar-baz')).to.equal('fooBarBaz');
+        });
+    });
+
+    describe('createFileNode()', () => {
+        const options = {
```
asmblah: This object should probably be recreated and assigned (so a `let` and not a `const`, ofc) inside a `beforeEach(...)`, in case the code accidentally modifies the object, in which case it could affect subsequent tests

tests/js/web/DropZoneInstanceManagerTest.js
```diff
@@ -0,0 +1,206 @@
...
+        });
+
+        it('should return an instance from an id', () => {
+            instanceManager.instances = mockInstances;
+            expect(instanceManager.getInstance(1)).to.deep.equal(mockInstances[1]);
+        });
+    });
+
+    describe('addFiles()', () => {
+        const mockInstance = {
```
asmblah: As above, this object should probably be recreated for each test in a `beforeEach(...)`

docs/_components/dropzone.md
```diff
@@ -0,0 +1,638 @@
...
+
+Any option from the reference above that is prefixed with `data-dropzone-` can be initialised using the twig helper (or HTML when in passive mode) or through the DropZone's `init(...)` method as a camel case version of the attribute.
+
+---
+
+#### Max Files
+
+```
+@param {String|Number}
+@deafult [5]
```
mikedevelops: I've made up my own syntax here, this is intended to be the default value of that option, maybe I should drop the pseudo JSDoc here all together if it's confusing.

asmblah: @mikedevelops it's fine, it's just spelt wrong 😄

js/DropZone/DropZoneComponent.js
```diff
@@ -0,0 +1,564 @@
...
+            // persist the front-end validation, which is essential when using an
+            // associated file input
+            if (input.value) {
+                this.instanceManager.addFiles(files, id, { persist: true });
+            }
+
+            // reset input node value, this will ensure our change event
+            // fires each time we use the browse files functionality - even
+            // if we try to add an identical value
+            input.value = '';
```
mikedevelops: It's only IE9 that kicks off about that (I believe), in which case we don't need to reset the input as it doesn't support any enriched features. I could make that clearer by putting this in the `if (this.supported)` condition.

js/DropZone/DropZone.js
```diff
@@ -0,0 +1,289 @@
...
+            { valid, text }
+        );
+    }
+
+    /**
+     * Return all files in the store or a file at a specified index
+     * @param {Number} index
+     * @return {Array|Object}
+     */
+    getFiles (index = -1) {
```
mikedevelops: This a 100 times, I thought I was being clever here, but it's been a real pain! I might leave this as is for v1.

mikedevelops: Changed my mind on this, all of the `getFiles` methods are split down into `getFiles()` & `getFile(index)`

js/DropZone/DropZoneComponent.js
asmblah: I _think_ IE10 doesn't like it either, but I haven't tested it, just did a quick Google: https://github.com/jquery/jquery/issues/3454

that change of `type` attr and back again might be a good workaround

docs/_components/dropzone.md
```diff
@@ -0,0 +1,638 @@
...
+
+dropZoneComponent.init();
+```
+{% endraw %}
+
+---
+
+### Passive
+
+DropZone initiated in passive mode.
```
mikedevelops: I've moved the passive description up to here, makes much more sense in this section.

docs/_components/dropzone.md
mikedevelops: ah, fixed 😄

js/DropZone/DropZone.js
```diff
@@ -0,0 +1,289 @@
...
+    }
+
+    /**
+     * Validate and add files to the DropZone
+     * @param {Array} files
+     * @param {Object} meta
+     */
+    addFiles (files, meta = {}) {
+        const { valid, text } = this.validator.validate(files, this.files.length, this.size);
+        let processedFiles = [];
```
mikedevelops: Yep, not sure how that slipped through, probably didn't start life as an array 😄

js/DropZone/DropZoneComponent.js
mikedevelops: I like the easy win `type` change, gone with this :+1:

js/DropZone/DropZoneComponent.js
```diff
@@ -0,0 +1,564 @@
...
+     * @param {String} string
+     */
+    updateInfoState (id, string) {
+        const { info, input, node } = this.instanceManager.getInstance(id);
+        const options = this.optionsManager.getInstanceOptions(id);
+        let browse;
+
+        if (info) {
+            info.innerHTML = string;
+            browse = node.querySelector(`.${options.nodeClasses.browse}`);
```
mikedevelops: This is a good call, likely another "used to be a legit let but then got refactored to within an inch of it's life" scenarios 😄

js/DropZone/DropZoneComponent.js
```diff
@@ -0,0 +1,564 @@
...
+        }
+
+        // create file html string
+        files.forEach(file => {
+            fileNodeString += this.utils.createFileNode(file, options);
+        });
+
+        // update wrapper html
+        wrapper.innerHTML = fileNodeString;
+        [].slice.call(wrapper.querySelectorAll(`.${options.nodeClasses.close}`)).forEach(closeButton => {
```
mikedevelops:
```
[].forEach.call(document.querySelectorAll('div'))
VM111476:1 Uncaught TypeError: undefined is not a function
    at NodeList.forEach (<anonymous>)
    at <anonymous>:1:12
```

Doesn't seem to work, I would guess this is because `ForEach` doesn't return anything?

js/DropZone/DropZoneComponent.js
```diff
@@ -0,0 +1,564 @@
...
+            this.classManager.update(this.body, [instance.options.interactionClasses.dropZoneSuccess]);
+        } else {
+            // If a persist property has been set on the file, we will throw a
+            // validation error which will persist once the file is dropped, this
+            // helps us out when we need to throw an error when a user uses an associated
+            // native file input's "browse files"
+            let persist = false;
+            // a crude shim for Array.find, we just need to know if any of the files
+            // have the persist flag set to true
+            files.forEach(file => !persist && file.meta.persist ? persist = true : null);
```
mikedevelops: I've simplified this by ditching the find shim altogether in favour of `_.find(...)`

js/DropZone/DropZoneComponentValidationManager.js
```diff
@@ -0,0 +1,45 @@
+class DropZoneComponentValidation {
+    /**
+     * Remove validation node
+     * @param {Element} node
+     * @param {String} selector
+     */
+    clear (node, selector) {
+        const validation = node.querySelector(`.${selector}`);
+
+        validation && validation.parentNode.removeChild(validation);
```
mikedevelops: I agree, I was trying too hard to be be schmancy

js/DropZone/DropZoneEventManager.js
```diff
@@ -0,0 +1,38 @@
...
+    add (node, event, handler) {
+        this.pool.push({ node, event, handler });
+        node.addEventListener(event, handler);
+    }
+
+    /**
+     * Remove all event listeners and return an empty pool
+     */
+    removeAll () {
+        this.pool = this.pool.reduce((pool, item) => {
```
mikedevelops: agreed, reducer really wasn't the right tool for the job here.

js/DropZone/DropZoneFileUtils.js
```diff
@@ -0,0 +1,56 @@
...
+    /**
+     * Format type for printing
+     * @param  {String} type
+     * @return {String}
+     */
+    getFileType (type) {
+        if (type.length) {
+            return type;
+        } else {
+            return 'application/file';
```
mikedevelops: This is a "presentational" MIME, so for files we can't detect this was going to be string printed on the front end, I'm thinking maybe this should just return an empty string an not display a type for types we can't match.

js/DropZone/DropZoneOptionsManager.js
```diff
@@ -0,0 +1,89 @@
...
+    buildValidatorOptions (options) {
+        const validatorOptions = { validationText: {} };
+
+        return Object.keys(options).reduce((validatorOptions, option) => {
+            switch (option) {
+                case 'validationMaxFiles':
+                case 'validationWhitelist':
+                case 'validationMaxSize':
+                case 'validationUnknown':
+                    // todo - this could do with perhaps being re-thought about, this seems overly complicated
```
mikedevelops: Removed 😄 , it's complicated but it'll do for now

js/DropZone/DropZoneValidationUtils.js
```diff
@@ -0,0 +1,65 @@
...
+     * @param {number} max
+     * @returns {boolean} valid
+     */
+    validateCount (count, max) {
+        return count <= max;
+    }
+
+    /**
+     * Validate against max file size
+     * @param {number} count
```
mikedevelops: Do you mean `Number` over `number` here?

js/DropZone/DropZoneValidatorDispatcher.js
mikedevelops: yep good spot, Nikos pointed this one out also. Initially I was weary of needing the index, but I've been through and it's not required.

mikedevelops: Although saying that, annoyingly `Symbol` doesn't have out-of-box support in our PhantomJS env.

tests/js/web/DropZoneComponentTest.js
mikedevelops: Trying to stub the `HTML` element proved more trouble than it was worth, tiny refactor to have the plugin take the `body` as it's root node so we can stub it with a `div`

js/DropZone/DropZoneValidationUtils.js
asmblah: Nope, I meant renaming `count` to `size` - but now I know it was "intentional"...

js/DropZone/DropZoneComponent.js
asmblah: As discussed, `[].forEach.call(document.querySelectorAll('body'), function (item) {...})` does work but isn't worth changing as we slice away in other places.

js/DropZone/DropZone.js
```diff
@@ -89,7 +92,7 @@ export default class DropZone {
      */
     addFiles (files, meta = {}) {
         const { valid, text } = this.validator.validate(files, this.files.length, this.size);
-        let processedFiles = [];
+        const processedFiles = [];
```
nvasileiadis: If processedFiles is a const now, why you don't put it in the line above with a comma?

mikedevelops: Personal preference, I don't like joined variable definitions (there will be lots of examples of this in the PR), for me joined declarations are more difficult to read and change.

PR #542, opened 2017-04-24
[CMS-4534] Add Iterated IDs to Primary Nav Items

nvasileiadis:
This structural change is necessary for the implementation of the Skip Nav Link Accessibility feature of CMS. As IDs are used for same page navigation via Tabs.

### DIFF
--- a/views/pulsar/v2/helpers/nav.html.twig
+++ b/views/pulsar/v2/helpers/nav.html.twig
@@ -15,7 +15,7 @@
 
             {% if options.data is iterable %}
                 {% for label, options in options.data %}
-                    {{ nav.nav_item(options, loop.index) }}
+                    {{ nav.nav_item(options) }}
                 {% endfor %}
             {% endif %}
         </ul>
@@ -34,7 +34,7 @@
             <ul class="nav-items">
                 {% for menu_blocks in options.items %}
                     {% for item_options in menu_blocks %}
-                        {{ nav.nav_item(options, loop.index) }}
+                        {{ nav.nav_item(item_options) }}
                     {% endfor %}
                     {% if not loop.last %}
                         <li class="nav-item nav-item__divider">
@@ -56,12 +56,12 @@
 {% endmacro %}
 
 
-{% macro nav_item(options, index) %}
+{% macro nav_item(options) %}
 {% spaceless %}
     {% import '@pulsar/pulsar/v2/helpers/html.html.twig' as html %}
     {% import '@pulsar/pulsar/v2/helpers/util.html.twig' as util %}
 
-    <li id="nav-item-{{ index }}"
+    <li
         {{
             attributes(options
                 |exclude('data-toggle href icon')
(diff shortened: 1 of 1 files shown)


views/pulsar/v2/helpers/nav.html.twig
```diff
@@ -56,12 +56,12 @@
...
 
 
-{% macro nav_item(options) %}
+{% macro nav_item(options, index) %}
 {% spaceless %}
     {% import '@pulsar/pulsar/v2/helpers/html.html.twig' as html %}
     {% import '@pulsar/pulsar/v2/helpers/util.html.twig' as util %}
 
-    <li
+    <li id="nav-item-{{ index }}"
```
mikedevelops: Could we get caught out here if there are multiple instances of the nav? (not sure if that is a possibility), example - 2 nav items with the id `nav-item-0`. Do we need to namespace instances of the primary nav?

nvasileiadis: As far as I am aware we have only one instance of the Primary Navigation every time. No, we don't need a specific namespace but just giving it an id like "01" felt not descriptive enough.

mikedevelops: Sounds good in that case, sorry by name spacing nav instances I meant... have the `<li>`s id something like `id="nav-{{ navInstanceIndex}}-item-{{ itemIndex}}` to get around multiple instances - but if multiple instances are not a concern, ignore me!

nvasileiadis: Cheers @mikedevelops !

Stanton: There's two separate loops that use `nav_item` and I _think_ each one will kick off a new index, so straight away we'll have duplicate integers in those IDs like @mikedevelops suspected.

There's a more fundamental choice to be made first though... is the 'skip nav link' accessibility thing a CMS feature only? If so, I don't believe we should make this change here, the `nav_data` object that the CMS generates in it's page controller should be updated to pass an `id` attribute with their `$this->navigation` object, this will then be added to the nav item `li` by the attribute parser on L:66. That way, we're not generating redundant IDs for the other products.

More importantly, if anyone tries to pass an ID, or the controller is updated to do it, your change here on L:64 wouldn't be merged with the other ID and would now result in:

```
<li id="nav-item-0" id="foo">
```

**But**, if you plan on adding this skip nav link to Pulsar core, then we can go ahead with the change, but we'll need to find a way to:

* Avoid duplicated ID attributes (`id="a" id="b"`)
* Avoid duplicated ID values from `loop.index`
* Merge user-supplied IDs with the index ID

There's maybe something we could do with `loop.parent`, but that might only work with nested loops so don't quote me on that... first step would be adding some failing unit tests which we can work through tomorrow 😀

nvasileiadis:
A less complex solution was found. There fore this Pull Request will be closed without merging.

PR #546, opened 2017-04-27
Added range slider styles for html range inputs

jamesjacobs:
Currently used in CXM for image zoom.
slider

Haven't added a helper yet as styles are used via `@extend` to style external lib.

Tested in:

- [x] IE Edge
- [x] IE 11
- [x] IE 10
- [x] Safari
- [x] Chrome
- [x] Firefox

Falls back to standard file input in < IE10 so no old IE styles.


Stanton:
While you've not added a helper here, it feels weird that the class/styles aren't documented... worth adding a doc page in the forms helper section with some examples, and add a ticket to create a helper soon?

jamesjacobs:
@Stanton Originally it was much DRYer, however, combining the vendor specific track and thumb selectors broke the styles. Will take another look and add docs.

jamesjacobs:
You can't combine vendor specific pseudo-elements/classes, see https://www.w3.org/TR/CSS21/syndata.html#rule-sets, specifically

> The selector (see also the section on selectors) consists of everything up to (but not including) the first left curly brace ({). A selector always goes together with a declaration block. When a user agent cannot parse the selector (i.e., it is not valid CSS 2.1), it must ignore the selector and the following declaration block (if any) as well.

Stanton:
Is this one still still waiting on stuff? or ready to merge?

jamesjacobs:
Just making a small change then needs to go back through testing.

PR #552, opened 2017-05-08
Add Primary Nav Item Height Adjustment

nvasileiadis:
This commit checks the content of the label inside a Primary Nav Item and when it is too big changes the height and the padding to look similar to the other items.

### DIFF
--- a/stylesheets/_component.navigation.scss
+++ b/stylesheets/_component.navigation.scss
@@ -244,7 +244,7 @@ $nav-color-dark:  darken(color(jadu-blue, dark), 5%) !default;
         font-size: .75em;
         line-height: 1em;
         max-width: $nav-primary-width;
-        min-height: 72px;
+        min-height: 72px; /* This is the optimal height for normal nav links with icon and one line text */
         min-width: $nav-primary-width;
         padding: 16px 5px;
         position: relative;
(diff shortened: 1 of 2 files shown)


nvasileiadis:
eventsformsmappings

Stanton:
Is JS the best way to address this or could the nav items simply have a sensible CSS `min-height` and then stretch vertically to accommodate large titles?

nvasileiadis:
Directly related to SPC-29605 [(https://jadultd.atlassian.net/browse/SPC-29605)]

nvasileiadis:
Absolutely right @Stanton totally forgot that way

stylesheets/_component.navigation.scss
```diff
@@ -239,13 +239,14 @@ $nav-color-dark:  darken(color(jadu-blue, dark), 5%) !default;
     .nav-link {
         color: lighten($nav-color-light, 40);
         display: block;
-        height: 72px;
+        height: 100%;
         font-family: $font-family-regular;
         font-size: .75em;
         line-height: 1em;
         max-width: $nav-primary-width;
+        min-height: 72px;
```
mikedevelops: Do we think it'd be worth adding a "magic number" comment here? This would help the next person to arrive at this block understand how you got to the value 72px. https://sass-guidelin.es/#magic-numbers

nvasileiadis: Done

PR #554, opened 2017-05-15
Allow the empty table background image path to be overidden

Stanton:
So that XFP can point the `$image-path` var to the vendors directory to wire up the image


stylesheets/_utility.utilities.scss
```diff
@@ -78,7 +78,7 @@ p.success {
 
 .no-results {
     background-color: color(gray, off-white);
-    background-image: url('../images/stripes.png');
+    background-image: url($image-path + 'stripes.png');
```
jamesjacobs: Has a default value for `$image-path` been set?

Stanton: It's a config variable, everything would break without one

Issue #558, opened 2017-05-19
Add ability to prevent select2s from being initialised


Stanton:
Currently, any select2 with the appropriate classes is automatically initialised by PulsarFormComponent, in order to provide non-standard options, like ajax methods, we need a way to prevent the automatic behaviour.

PR #559, opened 2017-05-19
Add data attribute to prevent select2s from being automatically initialised

Stanton:
Example:

```twig
{{
    form.select2({
        'label': 'Pick a colour',
        'id': 'foo',
        'data-init': 'false',
        ...
```


jamesjacobs:
Closes #558

PR #564, opened 2017-05-25
Allow block list items to be visually disabled

### LINKED ISSUE #563: Block list items can't be disabled
> James Manley [4:14 PM] 
> Is it possible to disable a block list item?

Need a way to visually distinguish disabled items within a block list

Stanton:
**Helper usage:**
```
{{
    html.block_list({
        'items': [
            {'label': 'Foo', 'href': '#foo'},
            {'label': 'Bar', 'href': '#bar', 'class': 'is-disabled'},
            {'label': 'Baz', 'href': '#baz'}
        ]
    })
}}
```

**HTML usage:**
```
<ul class="block-list">
    <li class="block-list__item">One</li> 
    <li class="block-list__item is-disabled">Two</li>
    <li class="block-list__item">Three</li>
</ul>
```

**Expected visuals:**
screen shot 2017-05-25 at 09 09 16

Closes #563


stylesheets/_component.block-list.scss
```diff
@@ -101,6 +101,10 @@ $width: 500px;
     }
 }
 
+.block-list__item.is-disabled {
```
mikedevelops: Not sure if this is a relevant comment, but this feels half BEM half not (which may not be a problem) - I'd expect this to add a modifier a long the lines of `block-list__item--disabled`

Stanton: It's a good question which warrants explaining :) in Pulsar we use `is-disabled` as a common modifier class that is supported across many components and helpers. 

Having `block-list__item--disabled` puts a lot of burden on the developer wanting to quickly put together a Pulsar UI, needing to know that they're using a block list, that the thing is a block list item, and that we're using BEM to internally organise styles. It'd need documenting, and almost always have to be looked up when required.

Far, far easier that someone implementing a Pulsar UI should get predictable results by adding `is-disabled` to anything, whether it's a link, a button, or a more complex component like a form control.

Similar modifiers would be `is-active` `is-inactive` `is-open` `is-closed` etc.

Bit more info here: https://jadu.github.io/pulsar/styleguides/state/

PR #568, opened 2017-06-12
New simpler grid system

Stanton:
Documentation here: https://jadu.github.io/pulsar/layout/grid/

Existing Susy implementation has been deprecated as it's still in use for CXM,


stylesheets/_layout.sidebar.scss
```diff
@@ -2,7 +2,7 @@
 	display: none;
 	float: none;
 	height: 100%;
-	padding: 0 0 0 40px;
+	// padding: 0 0 0 40px;
```
mikedevelops: I'm going to be _that_ guy again! If we don't need this line shall we remove it? If it ever needs switching back we can use git to grab the old value. There's a few more instances down below 👇

stylesheets/_structure.grid.scss
```diff
@@ -1,36 +1,72 @@
...
+    @include column($first);
+
+    @if $force == false {
         @include respond-min($screen-phone) {
-            @include span($i);
+            width: 100%;
         }
     }
+    @else {
+        width: ($one_col * ($num)) + ($gutter_pc * ($num - 1));
```
mikedevelops: Could `($one_col * ($num)) + ($gutter_pc * ($num - 1))` be hoisted into a variable at the start of the block ☝️, it could make this mixin easier to reason about if we reduce those duplicate calculations below maybe?

PR #572, opened 2017-06-28
Make form helpers behave on the grid

Stanton:
This change makes form helpers use the grid for their main layout, it also slightly refactors some of the previous grid work in order to avoid overly specific overrides.

Form helpers are arranged on the grid as follows:
* `control__label` = 3 columns
* `controls` = 9 columns

grid-3 copy

The actual input inside of the `controls` block varies in width, text style inputs (vs checkboxes etc) use 4 columns by default so visually, a regular text input would be 7 columns wide (3+4) out of a potential 12 columns. 

That being said, 4 columns default is **wider** than the previous regular width of 300px, which may get raised by UAT/Test engineers as a bug, overall it's far more desirable having more responsive forms, so we'll just have to set expectations when pushing this release through the products.

# Main changes since last release

* Grid columns now require a `g-col` class, this was so that I could decouple the block level column styling from the width settings and allow us to easily give widths to other elements without also giving them column block styling and margins
* Form helpers need a form specific class to modify the width of the input (because using the normal grid class will modify the width of the entire `form__group` block (by design)
    e.g.: `.form__control--col-2` for 2 column input widths
* Added a new param to the `span()` mixin to specify the number of columns that will be used to calculate the % widths of a span, this is so that the `form__control--col-x` classes can correctly calculate widths for the inputs against a 9 column grid (where the first three were used for the label) instead of the default 12

# Potential regression issues

* I've (finally) removed the use of `:not(.checkbox):not(.radio):not(.file)` from all the selectors I could find as it's caused us specificity headaches since it was included
* Some elements that used margin to line up with the controls (like help blocks and `.form__group--indent`) now use the `push()` grid mixin instead of margin 
* Previously, inputs were set to pixel widths, now they are all % based and responsive. This could cause unintended layout issues for UIs that expect them to be specific widths
* The existing width modifiers (e.g. `.form__group--large`) are still supported, but are mapped to specific column widths instead of the old px widths, this *could* be reverted if required, but I kinda want to see how testing pans out... The Lexicon has been updated to show examples using both the grid and 'legacy' classes


Stanton:
Updated the ticket description 💃

stylesheets/_structure.forms.scss
```diff
@@ -6,24 +6,21 @@
...
     // the main label, usually placed in the left hand column
     > .control__label {
-        width: $label-column-width;
+        @include span($form-label-cols);
         word-wrap: break-word;
     }
 
     .help-block {
-        margin-top: .675em;
+        margin: .675em 0 0 1%;
```
mikedevelops: Can we have a magic number comment here `.675em 0 0 1%` looks pretty gnarly

stylesheets/_structure.forms.scss
```diff
@@ -85,29 +82,25 @@ $control-column-width: 75%;
...
+        @include push($form-label-cols);
+    }
 }
 
 // control group inside first column (normally used for control labels)
 .controls:first-child {
+    @include span($form-label-cols);
     display: table;
-    width: $label-column-width;
+    // width: $label-column-width;
```
mikedevelops: Let's remove this line if we don't need it

Issue #571, opened 2017-06-27
CSS Hunger Games


Stanton:
In the blue corner, we have `pulsar.css`, weighing in at a hefty 0.51mb uncompressed.

In the red corner, Team Pulsar!

Our challenge through July is to optimise as much of Pulsar's CSS as we can.

This is not an individual competition, you are encouraged to collaborate as much as possible and to pair program to resolve particularly tricky areas. We're all in this together, but there may be prizes for individual outstanding achievement ;)

A starter for 10 would be the widespread (mis)use of `@extend`.

Name your branches `hunger-games/your-branch` and point pull requests to the `hunger-games/develop` branch instead of main develop.

[CSS Stats](http://cssstats.com/stats?link=https%3A%2F%2Fjadu.github.io%2Fpulsar%2Fcss%2Fpulsar.css) 

<img width="1067" alt="screen shot 2017-06-27 at 21 18 10" src="https://user-images.githubusercontent.com/18653/27608127-2d372c34-5b7e-11e7-8c5b-3c17f20360a9.png">

![](http://25.media.tumblr.com/506dadbf234118f464ba64efc6bfec3a/tumblr_mgmx1aYSZp1s37nm3o1_500.gif)

Stanton:
You can treat this ticket like a develop branch for all hunger games related PRs, there's also a milestone that you should use to keep track of everything.

Stanton:
🏆 Longest selector award

`.form__button-group .form__control:checked + .filter-bar__add.control__label.remove-button:hover:not(:disabled):not(.is-disabled).btn--white `

Stanton:
```
┌─────────────────────────────────┬────────────────────────────────────────────────────────────────────┐
│ Published                       │ July 3, 2017 11:06 AM                                              │
├─────────────────────────────────┼────────────────────────────────────────────────────────────────────┤
│ Paths                           │ css/pulsar.css                                                     │
├─────────────────────────────────┼────────────────────────────────────────────────────────────────────┤
│ Style Sheets                    │ 1                                                                  │
├─────────────────────────────────┼────────────────────────────────────────────────────────────────────┤
│ Style Elements                  │ 0                                                                  │
├─────────────────────────────────┼────────────────────────────────────────────────────────────────────┤
│ Size                            │ 556.9KB                                                            │
├─────────────────────────────────┼────────────────────────────────────────────────────────────────────┤
│ Data URI Size                   │ 4.5KB                                                              │
[...]

PR #576, opened 2017-07-04
Split button states into separate mixins

Stanton:
I noticed that the `.filter-bar__add` button was extending the entire primary button only for hover/active/focus states, buttons are quite a complex set of styles to cater for the myriad variations so it's not too much of a widespread refactor (button styles are responsible for a large number of selectors).

This change adds a couple of new mixins to allow you to include only hover/active states when required.

It’s only a modest saving, but it’s a start!

```
$ stylestats css/pulsar.css -c .stylestatsrc
 StyleStats!
┌──────────────┬───────────────────┐
│ Size         │ 533.1KB (-23.3KB) │
├──────────────┼───────────────────┤
│ Rules        │ 3705 (-11)        │
├──────────────┼───────────────────┤
│ Selectors    │ 6870 (-289)       │
├──────────────┼───────────────────┤
│ Declarations │ 7720 (-1)         │
└──────────────┴───────────────────┘
```

#571


nvasileiadis:
Is there a specific reason the $shadow-color has been removed? There are three instances that could be used within the given code.

Stanton:
@nvasileiadis It felt like needless abstraction really, It's either `$shadow-color` or `darken(color($state), 10)` and both are mostly variables. Plus now I've split things over a few mixins it felt like being more explicit was more readable/understandable by others.

PR #595, opened 2017-07-26
Change reference from align-right to u-text-align-right

### LINKED ISSUE #549: Align right class inconsistency
`.align-right` works

`.u-align-right` doesn't

Should be mapped to `u-` namespace for consistency.

Stanton:
Ticket #549 incorrectly stated `u-align-right` instead of `u-text-align-right`, so I've updated the related view to use the `u-text-align-right` instead, but the issue with specificity remained.

To resolve, I've increased specificty of all `u-` utility classes by one step, open to discussion on this one, but it fixes the cause of the issue where `.u-text-align-right` was not able to override the text alignment set by `.table th`.

Related reading: https://csswizardry.com/2014/07/hacks-for-dealing-with-specificity/

Also meant I didn't have to make them all immutable with `!important`, though there are valid reasons why we'd want utility classes to be immutable 🤔 

Closes #549


Stanton:
Changed to immutable utility classes following team discussion. Re-review pls 🙏

PR #596, opened 2017-07-27
Drag & Drop disable browse files

mikedevelops:
An improvement to the DropZone API to allow toggling of the "Browse Files" functionality.
jul-27-2017 10-01-29

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -17,7 +17,7 @@
     "noempty": true,
     "nonew": true,
     "plusplus": false,
-    "strict": "global",
+    "strict": false,
     "trailing": true,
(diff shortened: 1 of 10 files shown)


Stanton:
Is the gif exampling disabling because only one file can be attached in this instance?

mikedevelops:
Yes @Stanton, disabling the browse functionality doesn't come out-of-the-box as such, this enhancement exposes methods to allow the integration code to implicitly disable and re-enable "Browse Files". 

So for this example in the integration code I'm disabling "Browse Files" once a file has been dropped using the `customDropZoneDrop` callback.

Issue #525, opened 2017-03-28
Masterswitch component adds invalid disabled attribute to links


jamesjacobs:
Removing this causes other issues such as preventDefault not firing.

Suggest it follows a similar pattern to the DisableUiComponent.

PR #598, opened 2017-07-27
Masterswitch now uses DisableUIComponent to disable the masterswitch contents

Stanton:
This allows us to use a common method of disabling different interactive elements within Puslar.

* Creates new `enable()` and `disable()` methods in DisableUIComponent
* Refactors `MasterSwitchComponent` to use the new methods
* Fixes an issue where the `switchOn/Off` methods were being called on both `click` and `change`
* Removes some tests from `MasterSwitchComponentTest` and assumes they methods called are (or will be) tested within `DisabledUIComponentTest`

- [x] Add tests for new DisableUiComponent methods

Addresses #525

### DIFF
--- a/js/DisableUiComponent.js
+++ b/js/DisableUiComponent.js
@@ -40,7 +40,6 @@ DisableUiComponent.prototype.disable = function (target) {
             .addClass('js-disabled u-cursor-not-allowed');
 
         // Wrap with disabled wrapper to visually disable
-        console.log('1');
         $this.wrap('<div class="u-ui-disabled"></div>');
     });
 };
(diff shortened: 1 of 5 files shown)


js/DisableUiComponent.js
```diff
@@ -35,11 +38,43 @@ DisableUiComponent.prototype.init = function () {
         // Disable links
         $this.find(LINK_ELEMENTS)
             .on('click', preventDefaultAndStopPropagation)
-            .addClass('u-cursor-not-allowed');
+            .addClass('js-disabled u-cursor-not-allowed');
 
         // Wrap with disabled wrapper to visually disable
+        console.log('1');
```
jamesjacobs: Sneaky log

Stanton: spotted :)

Stanton:
Not sure if CodeCov will post again, but the test coverage for these files is now back up to 💯 🎉

PR #604, opened 2017-08-02
"Actions" Dropdown Badges for Selected Table Rows

nvasileiadis:
- The "Actions" dropdown now has badges to show how many table rows are selected

actionsbadges01

- Selected table rows should have is-selected class applied to visually identify them as the selected rows

- It is now possible, to add the badge and the word of item/items like the original example:
   e.g. Delete (6) items

actionsbadges02


js/PulsarSortableComponent.js
```diff
@@ -114,6 +114,24 @@ PulsarSortableComponent.prototype.initTables = function () {
...
+        // toggle checkbox state if the checkbox was not the target of the event
+        if (!$target.is($checkbox)) {
+            $checkbox.prop('checked', !selected);
+        }
+
+        // toggle row selected class
+        $this.toggleClass('is-selected');
+
+        component.actionsBadge();
+    });
```
mikedevelops: This block is fantastic 💃

js/PulsarSortableComponent.js
```diff
@@ -145,4 +163,39 @@ PulsarSortableComponent.prototype.updateOrder = function() {
...
 
+// Create or update "Actions" badge when a table row is selected
+PulsarSortableComponent.prototype.actionsBadge = function() {
+    var component = this,
+        checkedBoxesCount = component.$html.find('.has-badges table tr input.checkbox:checked').length,
+        badge = component.$html.find('.has-badges .btn__group.dropdown span.badge'),
+        itemWording = component.$html.find('.has-badges .btn__group.dropdown .item-wording'),
+        deleteOption = '.has-badges .dropdown__menu li:last-child a';
+
+    if (checkedBoxesCount > 0) {
```
mikedevelops: Very minor here, but we could tidy this up as `if (checkedBoxesCount) {`

js/PulsarSortableComponent.js
```diff
@@ -145,4 +163,39 @@ PulsarSortableComponent.prototype.updateOrder = function() {
...
+        // Remove previous badge
+        badge.remove();
+        // Add badge next to "Actions" dropdown text
+        $('<span class="badge">'+ checkedBoxesCount +'</span>').insertBefore('.has-badges .btn__group.dropdown span.caret');
+
+        if ($(deleteOption).text().indexOf("Delete") === 1) {
```
mikedevelops: Identifying this delete option by it's text seems fragile to me, could we not use an attribute or class here? You just know 12 months down the line someone will change that to "Remove"

js/PulsarSortableComponent.js
nvasileiadis: Done

js/PulsarSortableComponent.js
nvasileiadis: Done

js/PulsarSortableComponent.js
nvasileiadis: We have an else if, for the case it is zero, in that one. I think we should leave it as is

PR #610, opened 2017-08-04
Remove susy grid, fix CMS compilation issue

Stanton:
## Removal of susy dependency

Susy grid has been entirely removed to prevent the conflict of multiple mixins called `span` and `gutter`. The alternative option is to rename the pulsar mixins to prevent conflict, however Susy was already agreed to be deprecated earlier this year.

## Explicitly include config/grid mixins within responsive-utilities

CMS was hitting an issue where it complained about `span` not being a defined mixin, I think this is caused by multiple mixins becoming confused in combination with the include order.

| Script | Gets | Sets |
| --- | --- | --- |
| config.variables        |                                | screen width vars |
| mixins.responsive    | screen width vars | responsive mixins |
| structure.grid           | responsive mixins | grid mixins |
| component.select2 | all | select2 styles | 

```
>> Error: no mixin named span
>>
>>        Backtrace:
>>            vendor/jadu/pulsar/stylesheets/_component.select2.scss:11, in mixin `@content`
>>            vendor/jadu/pulsar/stylesheets/_mixin.media-queries.scss:7, in mixin `respond-min`
>>            vendor/jadu/pulsar/stylesheets/_component.select2.scss:10
>>         on line 11 of vendor/jadu/pulsar/stylesheets/_component.select2.scss
>> >>         @include span(4, false, true, 9, true);
>>    -----------------^
```

At the moment I suspect that for some reason, CMS is choking on the result of select2 attempting to use a combination of grid / responsive mixins together in a way that, well, just b0rks.

Patch fix for this is to include the config and grid, further investigation may reveal the actual reason this works in Pulsar but no in CMS...


Stanton:
Change is no longer required, #635 resolves this issue.

PR #620, opened 2017-08-16
Dropzone tooltip fix

Stanton:
Also released as tag `6.20.3` on `release/6.20.x` branch for CXM.


Stanton:
Closing so that the whole `release/6.20.x` branch can be merged into develop, which is easier.

Issue #605, opened 2017-08-02
Select2 width calculations taking place when hidden


mikedevelops:
With the current javascript tab system (in Lexicon) there is an issue with Select2 returning an element's percentage value as a pixel value. 

This is because it is unable to determine the element's parent's width to calculate the relative percentage value in pixels. It's unable to determine this on Select2s that are in tabs that are `display: none` as we can't get an elements physical dimensions when hidden (or at least Select2 doesn't - it can be, but probably shouldn't be,  hacked).

We need to hook into the tab plugin to have Select2s recalculate when a tab is made active, it'd be nice to only perform this calculation if the tab in question contains a select2 element.

PR #621, opened 2017-08-16
Init hidden select2 widths when their parent tab is opened

Stanton:
- [x] tabs
- [x] modals

* Given select2s are present on a hidden tab
* When the tab is shown
* Then the select2s are reinitialised so they can calculate their widths correctly

Before:
select2

After:
select2-after

Addresses #605


Stanton:
Need to fix this for modals too

screen_shot_2017-08-18_at_13 36 41_720

tests/js/web/PulsarFormComponentTest.js
```diff
@@ -80,25 +93,57 @@ describe('Pulsar Form Component - Select2 elements', function() {
...
+            this.pulsarForm.initSelect2 = sinon.stub();
             this.pulsarForm.init();
+
         });
 
         it('should call the select2 plugin', function() {
-            expect($.fn.select2).to.have.been.called;
+            expect(this.pulsarForm.initSelect2).to.have.been.called;
+        });
+
```
nvasileiadis: Empty line

PR #625, opened 2017-08-22
Tables with Dropdown Badges Documentation Entries

nvasileiadis:


docs/_components/tables.md
```diff
@@ -36,6 +36,33 @@ Table rows will accept the normal collection of state classes which will highlig
 * has-inverse
 * has-new
 
+## Tables with badges in "Actions" dropdown
+
+To add badges to the "Actions" dropdown when items are selected a class "has-badges" needs to be added in an element grouping both the dropdown and the table together.
```
Stanton: Would suggest noting that this could be a `div` or a `form` depending on what you need to do

docs/_components/tables.md
```diff
@@ -36,6 +36,33 @@ Table rows will accept the normal collection of state classes which will highlig
...
+
+![Screenshot](https://user-images.githubusercontent.com/1425876/28925357-92b0a9d2-785c-11e7-898f-7b0b986164e2.gif)
```
Stanton: Not sure it's best to host images here, can't guarantee their longevity... You could move this into `docs/assets/image_examples`

PR #632, opened 2017-09-04
Add projector-friendly colour palette

### LINKED ISSUE #631: Projector / High Contrast Mode
Pulsar is nicely tuned to the type of displays used on desktop, laptop and mobile devices, critical text and user interface elements meet current WCAG AA contrast criteria. However, the non critical UI elements can appear washed out or even invisible when using a projector to demonstrate the software.

Pulsar needs a modifier class which can invoke a higher contrast colour palette, Continuum products can then implement a user setting to toggle this class on a per-user account basis (making it a user preference rather than a system wide modification).

Stanton:
* Adds a string-replace task to the gruntfile, this takes `pulsar.scss` and replaces `base.palette` with `projector.palette`
* Replace task creates a new `pulsar-theme-projector.scss` file which will then be picked up by grunt-sass when building, and create new `pulsar-theme-projector.css` file which should be the same as `pulsar.css`, but with only the palette switched out
* Same behaviour happens for IE specific scss/css files

~~This can be tested by appending `?theme=projector` to any lexicon URL.~~

The projector theme currently affects the monochrome and state palettes, and provides darker versions of the lightest colours by a given `$darken-factor`, currently set to 20%.

Closes #631 

## Monochromes before/after
screen shot 2017-09-04 at 09 12 13
screen shot 2017-09-04 at 09 11 34

## States before/after
screen shot 2017-09-04 at 09 12 20
screen shot 2017-09-04 at 09 11 48


Stanton:
Not a great test as the projection isn't being thrown far and the room is quite dark because of the weather, but there's still a marked improvement when using the projector palette.

Before:
img_0210

After:
img_0209

Stanton:
Tagging as minor, doesn't break any existing behaviour within products but does require integrating with a product gruntfile and some development to implement.

PR #634, opened 2017-09-05
Fix CMS sass compilation issue

Stanton:
* Removes susy dependency
* Removes susy styles and mixins from deprecated stylesheet (@jamesjacobs, if CXM still depends on susy you'll need to move that dependency into CXM)

Voting open:

💥 Major
🔥  Minor
🚀 Patch


Stanton:
🚀

PR #638, opened 2017-09-11
616-masterswitch_disable_issues

### LINKED ISSUE #616: Masterswitch component doesn't disable active UI when disabled
Recent changes in masterswitch have broken some existing functionality.

When the layout is 

```
<form>
   <div class="masterswitch">
      ... content ...
   </div>
   <div class="form__actions">
   ... submit button ...
   </div>
</form>
```

The master switch does not disable the master switch contents when toggling from an enabled MS to a disabled MS.

Disabled (on load) to Enabled  - working
master switch working

Enabled (on load) to disabled - not working
master switch not working

Note - the form actions are outside of the masterswitch so the form can still be saved.

jamesjacobs:
As raised in #616 when a masterswitch UI was loaded with the toggle enabled, toggling to disabled did not disable the masterswitch content.

masterswitch fixes

Fixes #616


js/DisableUiComponent.js
```diff
@@ -70,7 +70,9 @@ DisableUiComponent.prototype.enable = function (target) {
             .removeClass('js-disabled u-cursor-not-allowed');
 
         // Remove wrapper which provides visually disabled styling
-        $this.unwrap('<div class="u-ui-disabled"></div>');
+        if ($this.parent().hasClass('u-ui-disabled')) {
```
Stanton: Do we need the check? I thought `unwrap()` failed silently?

> If an element's parent does not match the selector, the element won't be unwrapped.

jamesjacobs: This is needed as `.unwrap([selector])` is not supported in the version of jQuery currently in use.

PR #657, opened 2017-09-26
Form whitespace

Stanton:
# The problem

The whitespace added by non-helper markup breaks the layout:

screen shot 2017-09-26 at 12 05 11

# Proposed solution

Tweaking the gutter by .5% allows the presence of whitespace to go undetected:

screen shot 2017-09-26 at 12 04 54


stylesheets/_component.forms.scss
```diff
@@ -328,13 +330,15 @@ fieldset {
 // based on the remaining 9 column grid
 
     // default width
-    .control__label + .controls > .form__control:not(.checkbox) {
```
jamesjacobs: The `:not(.checkbox)` is needed here to fix checkbox widths and bg colours in FF and IE.

PR #663, opened 2017-10-03
Fixed checkbox bg and width issue in FF and IE

### LINKED ISSUE #662: Checkboxes have incorrect width and background colour
The issue exists in Firefox and IE11-9.

screen shot 2017-10-02 at 21 31 14

screen shot 2017-10-02 at 21 31 39

jamesjacobs:
Fixed the width issue in FF and IE (along with the bg colour issue).

Unfortunately due to the nested MQ generated by using the `col-span` mixin [here](https://github.com/jadu/pulsar/blob/662_checkbox-issues/stylesheets/_component.forms.scss#L332) when already inside [this MQ](https://github.com/jadu/pulsar/blob/662_checkbox-issues/stylesheets/_component.forms.scss#L307) the only option was to use a `not(.checkbox)` [here](https://github.com/jadu/pulsar/blob/662_checkbox-issues/stylesheets/_component.forms.scss#L331).

Previously this was generating this CSS
```
@media screen and (min-width: 600px) and (min-width: 480px) {
  .control__label + .controls > .form__control {
    width: 100%; } }
```

Which was more specific (by 1 MQ) than the width overrides (now removed).

Example of FF fix:
checkbox fix

The previous (non-working) overrides mentioned not using `not()` due to specificity so this fix may need testing in product.

Fixes #662

### DIFF
--- a/stylesheets/_component.forms.scss
+++ b/stylesheets/_component.forms.scss
@@ -90,6 +90,7 @@ fieldset {
         height: auto;
         padding: 0;
         vertical-align: baseline;
+        width: auto;
 
         @include ie-lte(9) {
             border: 0;
@@ -327,7 +328,7 @@ fieldset {
 // based on the remaining 9 column grid
 
     // default width
-    .control__label + .controls > .form__control {
+    .control__label + .controls > .form__control:not(.checkbox) {
         @include col-span(4, false, false, 9);
     }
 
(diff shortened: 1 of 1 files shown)


Stanton:
if this is affecting checkboxes enough to require `not(.checkbox)`, are radio inputs affected?

jamesjacobs:
@Stanton No. Radios are not affected by that selector as they are wrapped in `control__label`. I couldn't think of a situation where you would have a radio without a wrapping label unless I'm missing something?

Stanton:
@jamesjacobs No worries, I just remembered that in in the past we had `:not(.radio):not(.checkbox)` and wondered if there was an issue there too.

Issue #666, opened 2017-10-06
Move away from Bower


Stanton:
https://bower.io/blog/

Packages currently managed by Bower will be moved under the responsibility of NPM where possible.

mikedevelops:
My 2 cents...

One problem is once this is removed how we install Pulsar's "runtime" or "develop" dependencies, a (non ideal) solution is to keep the "lib" / "bower_components" directory committed, however I think this should be removed as part of this ticket. It isn't ideal having these committed from an architecture point of view, as well as not ideal we are opting out of patch and minor version bumps.

The reason this is a problem is because of how Jadu products currently consume Pulsar, they depend on the bower components being committed when Pulsar is installed with composer, when ideally they should be fetched with NPM (or some other trendy package manager) at the point Pulsar is installed / updated.

Having used both, I personally vote 👎  to yarn, as NPM v5 has a ["NPM 5 has stronger guarantees across versions and has a stronger deterministic lockfile"](https://yarnpkg.com/blog/2017/05/31/determinism/). Although admittedly we've had some teething issues with NPM v5, they are releasing patches monthly and they _seem_ to be clearing up.

Stanton:
@mikedevelops are we in a position to use NPM5 today, or are there any known blockers preventing us from starting to put together a PR?

mikedevelops:
@Stanton I can't foresee any known blockers, in fact, technically we are already using NPM v5, I am anyway, we already have a `package-lock.json` in the repo, so no work required on that front.

Stanton:
I'm hoping the ticket number isn't a sign 🙈

PR #668, opened 2017-10-09
First pass at de-bowering

Stanton:
Addresses #666 

| ---- | ---- |
| require changes to `main.js` | 🛑 Yes |
| require changes to `index.js` | 🛑 Yes |
| require changes to `gruntfile.js` | 🛑 Yes |
| require changes to `package.json` (not including dev dependencies) | 🛑 Yes |
| require changes to `base.html.twig` (or other core views like header/footer) | 🛑 Yes |
| require new arguments to be passed to a js component | 🛑 Yes |


Stanton:
Ready for review please, have repointed to merge into `pulsar7-develop` branch instead of `develop`

js/area/signin/signin.js
```diff
@@ -1,6 +1,6 @@
 var $ = require('jquery'),
-	placeholder = require('../../../libs/jquery-placeholder/jquery.placeholder'),
-	vide = require('../../../libs/vide/dist/jquery.vide.min');
+	placeholder = require('../../../node_modules/jquery-placeholder/jquery.placeholder'),
```
mikedevelops: Now these are living in `node_modules` we shouldn't need to specify the full relative path – there's nothing wrong with doing so, but for convention's sake shall we... 
 ```placeholder = require(jquery-placeholder/jquery.placeholder')```
or maybe even...
 ```placeholder = require(jquery-placeholder')```
now we aren't using bower packages have the ability to specify an entry point in their `package.json`s, so we can afford to be less specific when resolving their files (not the case for all though).

Stanton:
@mikedevelops Requested changes implemented

js/PulsarFormComponent.js
```diff
@@ -2,9 +2,9 @@
 
 var $ = require('jquery');
 
-require('../libs/pikaday/plugins/pikaday.jquery');
-require('../libs/select2/dist/js/select2.min');
-require('../libs/spectrum/spectrum');
+require('../node_modules/pikaday/plugins/pikaday.jquery');
```
mikedevelops: I'm fairly certain this will do `'pikaday/plugins/pikaday.jquery'`, webpack will automatically attempt to resolve require statements to the nearest `node_modules` folder recursively (in an upwards direction, not sure if "recursively" can mean in an upwards and downwards direction)

js/PulsarSortableComponent.js
```diff
@@ -4,7 +4,10 @@
 
 var $ = require('jquery');
 
-require('../libs/jquery-ui/jquery-ui.min');
+require('jquery-ui');
+
+// var mouse = require('jquery-ui/ui/widgets/mouse'),
+//     sortable = require('jquery-ui/ui/widgets/sortable');
```
mikedevelops: I think we should avoid committing commented out code, we can always restore them later if they are needed.

js/index.js
```diff
@@ -64,5 +52,5 @@ module.exports = {
     PulsarSortableComponent,
     SignInComponent,
     history,
-    svgeezy
+    svgeezy: window.svgeezy
```
mikedevelops: I'd expect `svgeezy` to work here (I could be wrong), my thinking is 

```{ svgeezy } === { svgeezy: svgeezy } === { svgeezy: window.svgeezy }``` 🤔

js/PulsarFormComponent.js
mikedevelops: I wikiapedia'd it, I think recursion applies to both directions 🍰

PR #675, opened 2017-10-17
658 sticky scroll bar

### LINKED ISSUE #658: Add support for a floating horizontal scroll bar to horizontal tables
When a horizontal scroll table has a large amount of content and a large number of columns the user has to scroll down, scroll across to then scroll up again to view the top rows.

If the horizontal scroll bar floated at the bottom of the table as the page scrolled it would solve this issue.

jamesjacobs:
Horizontally scrolling tables now have a sticky "floating" scroll bar to aid with navigating large tables with many columns and rows.

sticky-scroll-bar

Tested in:

- [x] IE9
- [x] IE10
- [x] IE11
- [x] Edge
- [x] Chrome
- [x] Safari
- [x] Firefox 

Fixed #658


js/StickyScrollBarComponent.js
```diff
@@ -0,0 +1,126 @@
...
+        // Init the sticky scroll bar
+        this.$scroller
+            .addClass('hide')
+            .on('scroll', this.scrollScrollableElemtentByScrollBarWithContext);
+
+        // Call on load
+        this.updateStickyScrollBarWithContext();
+
+        // Call on scroll and window resize
+        this.$window.on('resize', this.updateStickyScrollBarWithContext);
```
Stanton: Any benefit separating these rather than `this.$window.on('resize scroll'...` ?

js/PulsarUIComponent.js
```diff
@@ -103,7 +106,8 @@ PulsarUIComponent.prototype.initDataTables = function () {
     });
 
     datatablesHorizontal.each(function() {
-        var $this = $(this);
+        var $this = $(this),
+            $tableParent;
```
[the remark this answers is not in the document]
jamesjacobs: Removed

Stanton:
Very minor point, double bottom border caused by bottom of scrollbar and the top of the footer.

screen shot 2017-10-17 at 15 57 22

Stanton:
Small sweet spot where the scrollbar doesn't appear (but probably should)

scroller

jamesjacobs:
Pushed a fix for the sweet spot issue

js/StickyScrollBarComponent.js
```diff
@@ -0,0 +1,138 @@
+class StickyScrollBarComponent {
+
+    /**
+     * StickyScrollBarComponent
+     * @constructor
+     * @param {jQuery} $rootWindow - jQuery wrapper of the window object
+     * @param {jQuery} $html - jQuery wrapper of the html node
+     */
+    constructor ($rootWindow, $html) {
+        this.$elementWithStickyScrollBar = {};
```
mikedevelops: Why do we initialise this value with an object literal?

js/StickyScrollBarComponent.js
```diff
@@ -0,0 +1,138 @@
...
+            throw new Error('an element to add the scroller to must be passed to StickyScrollBarComponent');
+        }
+
+        if (typeof this.$window === 'undefined' || !this.$window) {
+            throw new Error('window must be passed to StickyScrollBarComponent');
+        }
+
+        if (typeof this.$html === 'undefined' || !this.$html) {
+            throw new Error('$html must be passed to StickyScrollBarComponent');
+        }
```
mikedevelops: Nice, I like this, it's responsible 🥇

js/StickyScrollBarComponent.js
```diff
@@ -0,0 +1,138 @@
...
+        // Add the sticky scroll bar
+        this.$stickyScrollBar.appendTo(this.$elementWithStickyScrollBar);
+
+        // Init the sticky scroll bar
+        this.$stickyScrollBar
+            .addClass('u-display-none')
+            .on('scroll', this.scrollElementWithStickyScrollBarByScrollBarWithContext);
+
+        // Call on load
+        this.updateStickyScrollBarWithContext();
```
mikedevelops: _This_ (pun) is an instance where we would not need to call this method with context, calling this method inside the `init(...)` closure will bind it with the context of `init(...)` which in this instance will be what we want – the component itself.

js/StickyScrollBarComponent.js
```diff
@@ -0,0 +1,138 @@
...
+
+        // Call on scroll and window resize
+        this.$window.on('scroll resize', this.updateStickyScrollBarWithContext);
```
mikedevelops: Should we consider debouncing these event handlers ? If the FPS is fine (~60), then perhaps not worth it, however worth thinking about as these events will potentially fire 100(0)s of times a second.

tests/js/web/FixedScrollBarComponentTest.js
```diff
@@ -0,0 +1,158 @@
...
+	});
+
+	afterEach(() => {
+        $stylesheet.remove();
+        $body.remove();
+    });
+
+    describe('init()', () => {
+
+		it('should throw an error if $window isn\'t passed to the component', () => {
```
Stanton: As a sneaky sneaky thing, you could use the `’` character here instead of `\'`

⌥ + ⇧ + ]

js/StickyScrollBarComponent.js
```diff
@@ -0,0 +1,138 @@
...
+
+    /**
+     * Toggle visiblity of sticky scroll bar
+     */
+    showStickyScrollBar (option) {
+        if (option) {
+            this.$stickyScrollBar.removeClass('u-display-none');
+        } else {
+            this.$stickyScrollBar.addClass('u-display-none');
+        }
```
mikedevelops: A little picky maybe – jQuery gives us an argument to `toggleClass` that could reduce this footprint somewhat
```javascript
showStickyScrollBar (option) {
    this.$stickyScrollBar.toggleClass('u-display-none', option);
}
```

js/StickyScrollBarComponent.js
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    /**
+     * Update sticky scroll bar visibility, thumb position and width
+     */
+    updateStickyScrollBar () {
+        const top = this.$elementWithStickyScrollBar.offset().top;
```
mikedevelops: ES6 to the rescue 🤺 

```javascript
const { top } = this.$elementWithStickyScrollBar.offset()
```

js/StickyScrollBarComponent.js
```diff
@@ -0,0 +1,138 @@
...
+        const bottom = top + this.$elementWithStickyScrollBar.height();
```
mikedevelops: Really minor so ignore, but in the pursuit of locking the FPS as high as we can, we could reduce the browser "layout" or sometimes referred to as "reflow" by grouping these rect queries. jQuery will make 2 calls here to `element.getBoundingClientRect(...)` to get your `offset.top` and `height`, if we simplified this we could make 1 call to get both values...
```
const { top, height } = this.$elementWithStickyScrollBar[0].getBoundingClientRect();
```

I'm not suggesting this is worth changing as [premature optimisation is the root of all evil](https://cdb.reacttraining.com/react-inline-functions-and-performance-bdff784f5578), but something I like to keep at the back of my mind, in terms of expensive DOM operations.

js/StickyScrollBarComponent.js
```diff
@@ -0,0 +1,138 @@
...
+        // Allow for footer
+        if (this.$html.find('.footer').css('position') === 'fixed') {
+            viewportBottom = this.$window.scrollTop() + this.$window.height() - this.$html.find('.footer').outerHeight();
+        } else {
+            viewportBottom = this.$window.scrollTop() + this.$window.height();
+        }
+
+        // Check if the $elementWithStickyScrollBar is visible but bottom is outside of viewport
+        if (top + topOffset < viewportBottom && bottom > viewportBottom) {
+
```
mikedevelops: empty line

tests/js/web/FixedScrollBarComponentTest.js
```diff
@@ -0,0 +1,158 @@
...
+
+			stickyScrollBarComponent.init($container);
+
+			expect($container.find('.sticky-scrollbar').hasClass('u-display-none')).to.be.false;
+		});
+
+	});
+
+	describe('When the container with the sticky scroll bar is scrolled', () => {
+
```
mikedevelops: Just a silly lint-esque comment, I don't think we should have newlines around the `describe` blocks, in the same way we wouldn't have them in other function closures in the codebase, but that's a personal one, so ignore if you disagree.

js/StickyScrollBarComponent.js
jamesjacobs: Good question... removing

js/StickyScrollBarComponent.js
jamesjacobs: 👍

js/StickyScrollBarComponent.js
jamesjacobs: Good spot, changing

js/StickyScrollBarComponent.js
jamesjacobs: Added for readability as the line below is also a comment. Personal preference, happy to remove.

js/StickyScrollBarComponent.js
jamesjacobs: 👍

PR #691, opened 2017-11-13
Create repeatable form component

mikedevelops:
## Acceptance criteria:

**The table**

- [x] should show an 'empty' message if no items have been added
- [x] should hide the empty message when an item exists in, or is added to the table
- [x] should allow items to be edited
- [x] should allow items to be removed

**The form**

- [x] should be hidden on initial load
- [x] should be shown when the 'add' button is pushed
- [x] should add a new row when the form 'add' button is pressed, and store the values in an appropriate way, ready for saving
- [x] should be hidden when the 'cancel' button is pressed
- [x] should be cleared when the 'cancel' button is pressed

**The add button**

- [x] should be labelled 'add' if no items exist in the table
- [x] should be labelled 'add another' if at least one item exists in the table
- [x] Should be disabled when the form is visible

**The component**

~should allow an optional `min` value, and trigger a validation error when the user has not added enough items~
~should allow an optional `max` value, and prevent the user from exceeding this number~
- [x] should disable the 'add' button if the `max` condition has been met

**Super Easter Egg Feature Bonus Round** 🥚 

- [x] should work __without__ Ajax

## Example markup:

```
<div class="form__group">
    <label class="control__label">Website accounts <span class="required-indicator" rel="tooltip" data-toggle="tooltips" title="" data-original-title="required">*</span></label>
    <div class="controls">
        <div class="table-container">
            <table class="table table--full">
                <thead>
                    <tr>
                        <th>Name</th>
                        <th>Username</th>
                    </tr>
                </thead>
                <tbody>
                    <tr>
                        <td colspan="2" class="muted">You have not added any website accounts</td>
                    </tr>
                </tbody>
                <tfoot>
                    <tr>
                        <td class="table__form" colspan="2">
                            {{
                                form.fieldset_start({
                                    'legend': 'Add Access Account'
                                })
                            }}
                            {{
                                form.text({
                                    'label': 'Actual name',
                                    'class': 'form__control-col--8',
                                    'required': true
                                })
                            }}
                            {{
                                form.text({
                                    'label': 'Email address',
                                    'class': 'form__control-col--8',
                                    'required': true
                                })
                            }}
                            {{
                                form.text({
                                    'label': 'Username',
                                    'class': 'form__control-col--8',
                                    'required': true
                                })
                            }}
                            {{
                                form.text({
                                    'label': 'Password',
                                    'class': 'form__control-col--8',
                                    'required': true
                                })
                            }}
                            {{
                                form.text({
                                    'label': 'Confirm password',
                                    'class': 'form__control-col--8',
                                    'required': true
                                })
                            }}

                            <div class="form__actions">
                                {{
                                    html.button({
                                        'label': 'Add',
                                        'class': 'btn--primary'
                                    })
                                }}
                                {{
                                    html.button({
                                        'label': 'Cancel',
                                        'class': 'btn--naked'
                                    })
                                }}
                            </div>
                        </td>
                    </tr>
                </tfoot>
            </table>
        </div>
        <button class="btn is-disabled" disabled>Add</button>
    </div>
</div>
```


Stanton:
@mikedevelops currently, if a field is set to `'required': true` the row is still added to the repeatable table, is that a conscious decision to leave that validation to the product, or should we prevent adding a row until all required fields have been completed?

js/Repeater/RepeaterComponent.js
```diff
@@ -0,0 +1,513 @@
...
+            placeholder: { query: '.repeater__empty-placeholder' },
+            editGroup: { query: `[${this.repeaterAttributes.editGroup}]` },
+            deleteGroup: { query: `[${this.repeaterAttributes.deleteGroup}]` },
+            preview: { query: `[${this.repeaterAttributes.previewId}]` },
+            previewUi: { query: `[${this.repeaterAttributes.previewUi}]` }
+        };
+
+        // Preview UI HTML
+        this.previewUiHTML = `
+            <td class="table__td table__td--align-right table-row__actions">
```
Stanton: `table__td` and `table__td--align-right` are not required, they're legacy styles, most of the table styles are currently on element selectors so the top level table _should_ just need `.table`

mikedevelops:
This is something I was expecting to have to look at, we will not be able to use the browsers native validation for required elements here (as the actual inputs that are sent in the request are hidden), so this will need to be handled by a custom validation service.

stylesheets/_component.repeater.scss
```diff
@@ -0,0 +1,5 @@
+.repeater__group {
```
Stanton: If this is the only style, would it be better to use a utility class?

views/pulsar/v2/helpers/form.html.twig
```diff
@@ -2107,3 +2107,87 @@ data-*        | string | Data attributes, eg: `'data-foo': 'bar'`
...
+                            </td>
+                        </tr>
+                        </tfoot>
+                    </table>
+                </div>
+                {{
+                    html.button({
+                        'data-repeater-add-group': 'true',
+                        'label': 'Add',
+                        'class': 'btn--primary repeater__add-group',
```
Stanton: shouldn't be a primary button

views/pulsar/v2/helpers/form.html.twig
```diff
@@ -2107,3 +2107,87 @@ data-*        | string | Data attributes, eg: `'data-foo': 'bar'`
...
+                        'type': 'button'
```
Stanton: Don't need type here, should default to button

views/lexicon/forms/repeatable.html.twig
```diff
@@ -0,0 +1,120 @@
...
+                    'checked': false,
+                    'id': 'switch',
+                    'label': 'Switch',
+                    'name': 'switch'
+                })
+            }}
+
+        {{ form.repeatable_end() }}
+
+        <input type="submit"/>
```
Stanton: This should be worked into the 'actions' option of the `form.end()` helper

views/pulsar/v2/helpers/form.html.twig
```diff
@@ -2107,3 +2107,87 @@ data-*        | string | Data attributes, eg: `'data-foo': 'bar'`
...
+        <div class="repeater__saved-data" style="display: none;"></div>
+        <div class="form__group">
+            {{
+                elem.label({
+                    'label': options.label,
+                    'required': options.required|default(false)
+                })
+            }}
+            <div class="controls">
+                <div class="table-container">
```
Stanton: This wrapper is added by [PulsarUIComponent](https://github.com/jadu/pulsar/blob/develop/js/PulsarUIComponent.js#L48) so could be omitted here if preferred

views/pulsar/v2/helpers/form.html.twig
```diff
@@ -2107,3 +2107,87 @@ data-*        | string | Data attributes, eg: `'data-foo': 'bar'`
...
+                    <table class="table table--full repeatable__table">
```
Stanton: You use `.repeater` but `.repeatable__` here?

views/pulsar/v2/helpers/form.html.twig
```diff
@@ -2107,3 +2107,87 @@ data-*        | string | Data attributes, eg: `'data-foo': 'bar'`
...
+                                {% for heading in options.headings %}
+                                    <th data-repeater-for-name="{{ heading.name }}">{{ heading.label }}</th>
+                                {% endfor %}
+                                <th rowspan="1" colspan="1"></th>
+                            </tr>
+                        </thead>
+                        <tbody class="repeater__preview-data">
+                            <tr class="repeater__empty-placeholder">
+                                <td colspan="{{ headingLength }}" class="muted">{{ options.empty }}</td>
+                                <td class="table__td table__td--align-right table-row__actions"></td>
```
Stanton: `table__*`classes not required

mikedevelops:
We could look into preventing the row being saved if the required fields have no values, I think we need to look at (maybe mockup) how this is communicated. I think this component should prevent a group being added if fields are not complete – then a separate pulsar validation component takes over and displays error messages

views/pulsar/v2/helpers/form.html.twig
```diff
@@ -2107,3 +2107,87 @@ data-*        | string | Data attributes, eg: `'data-foo': 'bar'`
...
+{% macro repeatable_end() %}
+{% import '@pulsar/pulsar/v2/helpers/html.html.twig' as html %}
+{% spaceless %}
+                                <div class="form__actions repeater__group-actions">
+                                    {{
+                                        html.button({
+                                            'data-repeater-save-group': 'true',
+                                            'label': 'Save',
+                                            'class': 'btn--primary repeater__save-group',
+                                            'type': 'button'
```
Stanton: `type` not required

views/pulsar/v2/helpers/form.html.twig
```diff
@@ -2107,3 +2107,87 @@ data-*        | string | Data attributes, eg: `'data-foo': 'bar'`
...
+                                        })
+                                    }}
+
+                                    {{
+                                        html.button({
+                                            'data-repeater-cancel-save': 'true',
+                                            'label': 'Cancel',
+                                            'class': 'btn--naked repeater__cancel-group',
+                                            'type': 'button'
```
Stanton: `type` not required

Stanton:
> We could look into preventing the row being saved if the required fields have no values, I think we need to look at (maybe mockup) how this is communicated. I think this component should prevent a group being added if fields are not complete – then a separate pulsar validation component takes over and displays error messages

Separate validation component sounds good, but also sounds out of scope for this feature, could we have the repeater toggle the relevant error classes on incomplete required fields if attempting to add the row, and then replace that when we come to add a validation component?

Stanton:
Presuming documentation will follow once any issues/changes are sweated out during pull review?

views/pulsar/v2/helpers/form.html.twig
mikedevelops: This is required, or at least I think it is. A button without an explicit type `button` will cause a form to submit as far as I know?

views/pulsar/v2/helpers/form.html.twig
mikedevelops: Oh I see, i didn't check the default type, although I set it explicitly because it was submitting my forms, I'll double check that default value.

mikedevelops:
That sounds like a good idea with the error class. Yep, docs will be my last stop. Cloning inputs needs some manual work which I'm doing at the minute, then I'll look at basic validating within the new groups.

js/Repeater/RepeaterPreviewService.js
```diff
@@ -0,0 +1,123 @@
...
+
+                preview.innerText = value;
+            }
+        });
+    }
+
+    /**
+     * Toggle disabled state of preview UI, if a preview ID is not passed in
+     * we'll disabled all preview UI elements
+     * @param previewId?
```
nvasileiadis: Why is the '?' needed here?

mikedevelops: That is the syntax in JSDOC for an optional function argument

stylesheets/_component.repeater.scss
mikedevelops: Could you give an example of this one, I'm not 100% sure of the usage of utility class here

views/pulsar/v2/helpers/form.html.twig
mikedevelops: Nice! Removed

views/pulsar/v2/helpers/form.html.twig
mikedevelops: What are your thoughts on what these actions should be in that case, I have a *Add* & *Save* (which is the same as update), both *cancel* actions seem to work with naked.

views/pulsar/v2/helpers/form.html.twig
mikedevelops: More investigation, we add `type="button"` as a default option when we define a button as an input, however if we want a button to be a `<button/>` we don't set `type="button"` as a default (which I think is correct as it might be odd if a button *didn't* submit a form by default), however in this case I don't want that behaviour.

views/pulsar/v2/helpers/form.html.twig
mikedevelops: see below 👇

views/pulsar/v2/helpers/form.html.twig
mikedevelops: see below 👇

stylesheets/_component.repeater.scss
Stanton: you could either use .u-display-none in the markup, or extend %u-display-none here if you really want the repeater__group class to exist, just felt a bit superfluous

views/pulsar/v2/helpers/form.html.twig
Stanton: 👍

views/pulsar/v2/helpers/form.html.twig
Stanton: You mean what style? Add & Save should just be regular grey buttons, the primary action would be higher up in the UI and be the main form save/submit action

stylesheets/_component.repeater.scss
mikedevelops: I had a go at doing this. I'm not sure `u-display-none` is a viable option in this instance as it is immutable – I need to be able to toggle the visibility of this group. I could switch from toggling the display property to toggling the `u-display-none` class, however I'm weary of coupling the two together.

PR #698, opened 2017-11-28
Remove Height 100% from nav-link

nvasileiadis:
This is the fix for Ealing (SUP-15045 / SUP008057)


Stanton:
May be worth taking this for a spin on the device lab

PR #705, opened 2017-12-06
Whitelist specific phrasing tags for the html.button helper

Stanton:
Restrict the allowable HTML elements within a `<button>` to the common phrasing tags we're likely to use.

### DIFF
--- a/views/pulsar/v2/helpers/html.html.twig
+++ b/views/pulsar/v2/helpers/html.html.twig
@@ -308,7 +308,7 @@ data-*  | string | Data attributes, eg: `'data-foo': 'bar'`
                 })
             )
         }}>
-            {{- options.label|default|raw -}}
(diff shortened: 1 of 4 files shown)


views/pulsar/v2/helpers/html.html.twig
```diff
@@ -308,7 +308,7 @@ data-*  | string | Data attributes, eg: `'data-foo': 'bar'`
                 })
             )
         }}>
-            {{- options.label|default|raw -}}
+            {{- options.label|default|striptags('<abbr><b><br><cite><code><dfn><em><i><img><kbd><q><small><span><strong><sub><sup>')|raw -}}
```
mikedevelops: I guess we manage this going forward for when the inevitable question comes in...

> How do I add a table as my label text

😄

PR #713, opened 2017-12-18
Fixes/changes for XFP Formbuilder branch

Stanton:
- Stop `col-span` mixins from creating fractional values
- Make sure icons within tables (e.g. edit/remove icons) have appropriate colours and hover effects
- Fix vertical alignment with nested form groups
- Restore expected behaviour of the `show_or` option for rule block helpers
- Explicitly require jQuery in the StickyScrollBarComponent


package.json
```diff
@@ -102,6 +102,6 @@
   "dependencies": {
     "jstree": "^3.3.3",
     "lodash": "^4.6.1",
-    "moment": "^2.15.2"
+    "moment": "^2.20.1"
   }
 }
```
mikedevelops: Living life on the edge with an npm update 😄

Stanton: hmm, I think I did that as part of trying to fix Travis when it broke, I'll revert this in favour of what's currently on develop

PR #728, opened 2018-01-11
Separate Lexicon Example Styles from Pulsar Styles

nvasileiadis:
Lexicon styles have been separated from pulsar and now they reside only in the dev. They are not included in the dist


Gruntfile.js
```diff
@@ -541,6 +554,7 @@ module.exports = function(grunt) {
         'scsslint',
         'replace',
         'sass:dev',
+        'sass:lexicon',
```
Stanton: You may want to also add this to the watch:css task

nvasileiadis: Fixed and added it to the scsslint task too

Issue #727, opened 2018-01-11
Colour mismatch with warning panels and flash messages


jamesjacobs:
Warning panels and warning flash message have a different BG colour. They should match.

![screen_shot_2018-01-11_at_10 53 53](https://user-images.githubusercontent.com/756393/34822581-5b7523f8-f6bf-11e7-8734-0e02e3685c86.png)

Check other colour combinations too. May as well double check AA contrast too whilst we're in there.

PR #731, opened 2018-01-15
Flash messages now use state colours

jamesjacobs:
Previously flash messages had different colours to states colours and therefore didn't match elements such as panels - see #727.

Flash messages now use state colours. Also I noticed flash message and panels were outputting some unneeded CSS for a handful of states that weren't applicable to them, these have now been filtered out from the respective loops. Therefore: 

- `panel--base`
- `panel--white`
- `panel--new`

No longer exist. These don't seem to have been used anywhere and were visually broken. Suspect these were created by mistake and as a by-product of adding `base`, `white` and `new` to `$state-colors`. Note the colours have not been removed from `$state-colors` as they are used elsewhere.

I've also removed the old flash colours form the palettes.

New colours (All meet AA due to state colour usage):
new

Old colours (some didn't meet AA):
old

### DIFF
--- a/stylesheets/_component.flash.scss
+++ b/stylesheets/_component.flash.scss
@@ -34,9 +34,8 @@
 // Default
 .flash,
 .flash--default {
-    background: color(inverse, flash);
+    background: color(inverse);
     color: color(inverse, alt);
-    text-shadow: 0 0 2px rgba(0, 0, 0, .5);
 
     a {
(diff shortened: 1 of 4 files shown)


Stanton:
Can we get some screenshots?

jamesjacobs:
@Stanton Just added

PR #733, opened 2018-01-15
Fix Reports Scheduling Styles

nvasileiadis:
reports-scheduling

Styles fixed


stylesheets/_component.dropdowns.scss
```diff
@@ -300,3 +300,23 @@
     background-color: darken(color(primary), 15%);
   }
 }
+
+.dropdown--medium ~ ul {
```
Stanton: Do we need the element selector here, can we find a way to not need it?

stylesheets/_component.dropdowns.scss
```diff
@@ -300,3 +300,23 @@
...
+    width: 300px;
+}
+
+.dropdown__title {
+    border-bottom: 1px solid color(border);
+    margin: 0;
+    padding-bottom: 4px;
+
+    a {
```
Stanton: We can probably achieve this through utility classes rather than overly specific element selectors, barring that, we'd be better some classes specific to the reporting UI

stylesheets/_component.dropdowns.scss
```diff
@@ -300,3 +300,23 @@
...
+.dropdown--medium ~ .dropdown__menu {
+    width: 300px;
+}
+
+.dropdown__title {
+    border-bottom: 1px solid color(border);
+    margin: 0;
+    padding-bottom: 4px;
+
+    span {
```
jamesjacobs: I'd consider using a dropdown specific class here rather than an element selector

PR #738, opened 2018-01-18
Add utility classes to remove top and bottom margins from an element

Stanton:
New classes `.u-no-margin-top` and `.u-no-margin-bottom`.

Also deprecate use of `.bottomless`.


stylesheets/_utility.utilities.scss
```diff
@@ -328,3 +323,13 @@ div.no-results {
 .u-width-auto {
     width: auto !important;
 }
+
+%u-no-margin-top,
+.u-no-margin-top {
```
jamesjacobs: Wondering we should name `u-margin-top-none` for consistency with other utility classes.

PR #741, opened 2018-01-22
Separate JS for Lexicon Examples from Pulsar

nvasileiadis:
With the changes of this PR, all the example/demo Lexicon JS code should be placed in js/lexicon and all the new JS files should be added in js/lexicon/lexicon-index.js in similar way it is done in the js/index.js


mikedevelops:
It looks as though we are bundling and committing the bundle to the repo, what's the thinking behind this?

I would have envisioned that the Lexicon-only bundle would live in the `dist` directory and be ignored along with the production and test bundles.

Could you also have the Lexicon templates reference this file in their markup, a simple `console.log('welcome to Lexicon')` might be a good test that everything is wired up.

Gruntfile.js
```diff
@@ -27,7 +28,8 @@ module.exports = function(grunt) {
             dist: {
                 files: {
                     'dist/js/bundle.js': ['js/index.js'],
-                    'dist/js/test.js': ['tests/js/web/index.js']
+                    'dist/js/test.js': ['tests/js/web/index.js'],
+                    'dist/js/lexicon-bundle.js': ['js/lexicon/lexicon-index.js']
```
mikedevelops: I think the bundle files should live in the dist directory, and not be committed to the repo.

js/lexicon/lexicon-index.js
```diff
@@ -0,0 +1 @@
+// Something is going to be added here soon
```
mikedevelops: Could you also add the bundled lexicon javascript to the lexicon templates, it might be nice to stick a console log in there so you can test that this file is included in lexicon, and has access to the `Pulsar` global variable.

PR #744, opened 2018-01-23
Table-detail UI pattern

jamesjacobs:
A new UI pattern which allows a user to view additional information relevant to a row whilst staying in context.

table-detail

Tested in:
- Chrome
- Safari
- Firefox
- IE11
- IE10
- IE9


js/TableDetailComponent.js
```diff
@@ -0,0 +1,97 @@
...
+            event.preventDefault();
+
+            let detailContent = $(event.currentTarget).closest('tr').data('table-detail-content');
+            let customDetailPanelTitle = $(event.currentTarget).closest('tr').data('table-detail-panel-custom-title');
+            
+            this.viewDetail(detailContent, customDetailPanelTitle);
+        });
+
+        // Close click listener
+        this.$detailPanel.find('[data-table-detail-close-panel]').on('click', (event) => {
```
mikedevelops: Super minor comment, which is open for opinion really. We _could_ omit anonymous arrow function brackets if it accepts only 1 argument (although I'm not sure what we really gain from that!) – also worth noting that we cannot do that in most cases when writing TypeScript, thought I'd mention it though:

```
node.addEventListener('click', event => {
    event.preventDefault();
});
```

tests/js/web/TableDetailComponentTest.js
```diff
@@ -0,0 +1,118 @@
...
+describe('TableDetailComponent', () => {
+	const clickEvent = $.Event('click');
+	let $html;
+	let $body;
+	let $tableHTML;
+	let tableDetailComponent;
+	let tableDetailComponentWithoutHTML;
+
+	beforeEach(() => {
+		$html = $('<html></html>');
```
mikedevelops: We need to be vigilant when using this pattern, now the browser environment test suite is fixed things like this will break it. If we replace the contents of the `<html/>` element it will prevent the tests from running. Does this _need_ to be a `<html/>` node, or could it live as a `<div/>` for the tests?

mikedevelops: alternatively maybe we could cache the HTML contents in the before each and restore it in the after each, I tend to favour the first option though.

js/TableDetailComponent.js
```diff
@@ -0,0 +1,97 @@
...
+
+        let $panelHtml = $(
+            '<div class="table-detail t-table-detail" data-table-detail-panel>' +
+            '   <div class="table-detail__header">' +
+            '       <button type="button" class="close table-detail__header-close" data-table-detail-close-panel aria-hidden="true">&times;</button>' +
+            '       <h4 class="table-detail__title" data-table-detail-panel-title>Detail</h4>' +
+            '   </div>' +
+            '   <div class="table-detail__body" data-table-detail-panel-body></div>' +
+            '</div>'
+        );
```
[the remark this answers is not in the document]
jamesjacobs: Ohh cool 👍

js/TableDetailComponent.js
```diff
@@ -0,0 +1,97 @@
...
+     */
+    constructor ($html) {
+        this.$html = $html;
+    }
+
+    /**
+     * Initialise
+     */
+    init () {
+        if (typeof this.$html === 'undefined' || !this.$html) {
```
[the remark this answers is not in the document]
jamesjacobs: Good spot!

Issue #745, opened 2018-01-24
Progress lists are not responsive


jamesjacobs:
At 350px.

![screen shot 2018-01-24 at 16 25 15](https://user-images.githubusercontent.com/756393/35343938-c2800edc-0123-11e8-8b45-3857b1868669.png)

PR #746, opened 2018-01-25
Responsive progress lists 📱

jamesjacobs:
Previously progress lists were not responsive. See #745

They are now 🎉

responsive progress lists


Stanton:
These seem really tall at mobile widths, reckon you could tweak the height to make them a bit more compact? Did you try centering the text at all?

PR #749, opened 2018-01-25
Programmatic Favicon notifications

mikedevelops:
A service that will update a users favicon.

I've added a "preset" method to the public API called `addCircleNotification(color, size)` which will cover most cases, as well as a `addCustomNotification((canvas, ctx) => {})` to allow users of the service to manipulate the favicon as they wish, however I suspect we'll end up expanding on the "preset" methods to cover more common cases.

The service "gracefully degrades", therefore environments that do not support any of it's features will silently fail with no adverse effect on the UX.

screen shot 2018-01-25 at 17 07 41


js/main.js
```diff
@@ -75,6 +76,28 @@
             supported: !lt10,
             showInputNode: lt10
         });
+
+        // Notifications
+        pulsar.faviconEditor.init();
+
+        // NOTIFICATIONS DEBUG, PLS REMOVE, THX
```
Stanton: To keep in Lexicon or to remove as per comment?

.babelrc
```diff
@@ -1,3 +1,3 @@
 {
-  "presets": ["es2015"]
+  "presets": ["es2015", "es2017"]
```
asmblah: I know we discussed it before, but as you're changing this anyway, should we use the `env` preset now?

js/Notifications/FaviconEditor.js
```diff
@@ -0,0 +1,187 @@
...
+        let data;
+
+        this.favicons.forEach(({ node }) => {
+            if (filterFileExtension(node.href, 'ico png') || filterDataEncodedURI(node.href)) {
+                setupPromises.push(this.setup(node));
+            }
+        });
+
+        try {
+            favicons = await Promise.all(setupPromises);
```
asmblah: Nice 👍

tests/js/web/Notifications/FaviconEditorTest.js
```diff
@@ -0,0 +1,115 @@
+const FaviconEditor = require('../../../../js/Notifications/FaviconEditor');
+const $ = require('jquery');
+const vanillaFavicon = require('./snapshots/favicon');
+const redCircleFavicon = require('./snapshots/faviconWithCricleNotification?radius=10&color=red');
```
asmblah: `faviconWithCricle` -> `faviconWithCircle` (and the file to match of course) 😄

tests/js/web/Notifications/FaviconEditorTest.js
```diff
@@ -0,0 +1,115 @@
...
+                    .catch(done);
+            });
+        });
+
+        describe('addCustomNotification', () => {
+            it('should draw a custom graphic onto an image', (done) => {
+                faviconEditor = new FaviconEditor($root[0]);
+                faviconEditor.init();
+
+                // Override serializer to return raw image data, we cannot use the default
```
asmblah: For consistency, should we also do this for the test above?

tests/js/web/Notifications/FaviconEditorTest.js
```diff
@@ -0,0 +1,115 @@
+const FaviconEditor = require('../../../../js/Notifications/FaviconEditor');
+const $ = require('jquery');
+const vanillaFavicon = require('./snapshots/favicon');
```
asmblah: Small thing, but this is more of a "fixture" than a snapshot as you wouldn't expect to need to update the fixture image very often, should this be in a folder called `fixtures` to differentiate them?

js/main.js
mikedevelops: I'm hoping to move this over to the Lexicon bundle once that is fixed.

tests/js/web/Notifications/FaviconEditorTest.js
mikedevelops: Yep I agree, I'll change this.

tests/js/web/Notifications/FaviconEditorTest.js
mikedevelops: Yeah good point, the compression / rendering doesn't seem to differ on the 2 environments when using primitive shapes – I'll change it though, probably best to remove compression from the equation.

.babelrc
mikedevelops: ... maaaaaybe 😸

tests/js/web/Notifications/FaviconEditorTest.js
```diff
@@ -0,0 +1,159 @@
...
+            // not asserting the actual href in this case, as it will be prepended by
+            // the host and port used by Phantom, I'm happy that if the value is not what
+            // I've changed it to then it has been restored
+            $html.find('link').each((index, element) => {
+                expect(element.href).to.not.equal('foo');
+            });
+        });
+    });
+
+    describe('serializer', () => {
```
asmblah: Is this page intentionally left blank?

docs/_components/favicon.md
```diff
@@ -0,0 +1,125 @@
...
+ * @param canvas {HTMLCanvasElement}
+ * @param context {CanvasRenderingContext2D}
+ * @param originalImage {HTMLImageElement} derived from the original favicon <link/> node
+ */
+function (canvas, context, originalImage) {
+    return canvas.toDataURL(`image/${getFileExtension(originalImage.src)}`);
+}
+
+// In this example we return the raw image data from our serializer, this return
+// value will be our promise resovle value when invoking update methods
```
asmblah: Typo in `resovle`

tests/js/web/Notifications/FaviconEditorTest.js
```diff
@@ -0,0 +1,219 @@
...
+
```
asmblah: Tests for this method are missing (unless they are somewhere else instead 😄)

PR #752, opened 2018-01-31
Stop disabled tabs being keyboard accessible

Stanton:
Fixes an issue where disabled tabs could still be navigated to by keyboard by programatically adding `tabindex="-1"` to disabled tabs.

## Before
disabled tabs1

## After
disabled tabs


views/pulsar/v2/helpers/tabs.html.twig
```diff
@@ -31,9 +31,11 @@
                 {% if tab.disabled is defined and tab.disabled == true %}
                     {% set href = '#' %}
                     {% set type = 'link' %}
+                    {% set tabindex = '-1' %}
                 {% else %}
                     {% set href = '#' ~ tab.id %}
                     {% set type = 'tab' %}
+                    {% set tabindex = '0' %}
```
jamesjacobs: Why do we need to add a tab index of 0 here? Could we not specify a tab index at all if the link is not disabled, as it then should follow sequential flow

tests/unit/Jadu/Pulsar/Twig/Macro/Fixtures/tabs/tabs_list-disabled.html
```diff
@@ -1,7 +1,7 @@
 <div class="nav-inline t-nav-inline">
     <ul class="nav-inline__list">
         <li class="nav-inline__item nav-inline__item--is-active is-active">
-            <a href="#" data-toggle="link" aria-disabled="true" class="nav-inline__link is-disabled">bar</a>
+            <a href="#" data-toggle="link" aria-disabled="true" tabindex="-1" class="nav-inline__link is-disabled">bar</a>
```
jamesjacobs: Pretty sure it does but just to double check - `is-disabled` will prevent mouse focus won't it? Otherwise, even though it has tabindex="-1" it could still be focused by mouse.

views/pulsar/v2/helpers/tabs.html.twig
jamesjacobs: Another couple of options (which you might have explored) would be:

1. rather than displaying an `a`, display a `span` with similar styles (removing the need for tabindex, css and JS)
2. remove the href completely, therefore the link cannot be focused by keyboard or mouse (no need for JS or tab index)

Stanton: > Why do we need to add a tab index of 0 here? Could we not specify a tab index at all if the link is not disabled, as it then should follow sequential flow

Not specifying the tabindex is the same as tabindex 0, so it felt more obvious for other developers to simply toggle the option between 1/0 as they need to.

> rather than displaying an a, display a span with similar styles (removing the need for tabindex, css and JS)

Using disabled links is fine imho and keeps things simpler for any developer needing to toggle disabled state easily, without having to rewrite markup on the fly.

> remove the href completely, therefore the link cannot be focused by keyboard or mouse (no need for JS or tab index)

Again, I'm thinking of other developers here, they'd have to save href state somehow if loading disabled links rather than simply toggling their disabled state.

tests/unit/Jadu/Pulsar/Twig/Macro/Fixtures/tabs/tabs_list-disabled.html
Stanton: Yes.

PR #753, opened 2018-02-05
Improve visual presentation of focused form fields

Stanton:
This improves the general obviousness of the user interface as well as improving keyboard navigation.
Also adds a new lexicon page with a sample of all form elements on.

## Before
focus-before

## After
(note alignment issues with checlbox/radios have been sorted since the gif was recorded)
focus-after

### DIFF
--- a/stylesheets/_component.forms.scss
+++ b/stylesheets/_component.forms.scss
@@ -527,8 +527,8 @@ fieldset {
 .form__control.radio {
     appearance: none;
     background-color: #fafafa;
-    border: 1px solid color(border);
-    box-shadow: none;
+    border: 0;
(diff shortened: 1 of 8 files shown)


stylesheets/_component.forms.scss
```diff
@@ -508,8 +520,60 @@ fieldset {
...
 .form__group.form-radio .form__control.radio {
-    margin-top: 17px;
-    width: auto;
+    // margin-top: 17px;
+}
+
+.form__control.checkbox,
+.form__control.radio {
+    appearance: none;
+    background-color: #fafafa;
```
nvasileiadis: Shouldn't the colour value be encoded as the other colours?

PR #757, opened 2018-02-05
Fix W3C HTML Validation Errors (2nd Iteration)

nvasileiadis:


libs/select2/src/js/select2/dropdown/search.js
```diff
@@ -11,7 +11,7 @@ define([
       '<span class="select2-search select2-search--dropdown">' +
         '<input class="select2-search__field" type="search" tabindex="-1"' +
         ' autocomplete="off" autocorrect="off" autocapitalize="off"' +
-        ' spellcheck="false" role="textbox" />' +
+        ' spellcheck="false" />' +
       '</span>'
     );
```
mikedevelops: Should we be editing 3rd-party library files? I think we need to be careful editing this sort of stuff, as the plugin likely considers this markup internal and could have logic bound to it being in a certain state.

nvasileiadis: Oops I missed that. Will revert

Stanton:
@nvasileiadis could you resolve the conflicts in this one, probably caused by me making sure radio examples always show a `checked` option.

PR #764, opened 2018-02-16
Increased the width of the main nav to allow longer words in nav links

jamesjacobs:
Before:

screen shot 2018-02-16 at 13 04 38

After:

screen shot 2018-02-16 at 13 02 41

Tested in:

- [x] IE9 
- [x] IE10 
- [x] IE11
- [x] Edge
- [x] Safari
- [x] Firefox
- [x] Chrome


stylesheets/_component.search.scss
```diff
@@ -11,7 +11,7 @@
     }
 
     @include respond-min($screen-desktop) {
-        left: 95px;
+        left: 105px;
```
mikedevelops: It seems to me like we should utilising variables here, if the search position is relative the nav width, can't we reference the `$nav-primary-width` and add left hand padding/margin value?

nvasileiadis:
It plays nicely in Edge too
fat-nav-edge

stylesheets/_component.search.scss
jamesjacobs: Good spot, pushed a fix

PR #776, opened 2018-03-08
Fix More Icon breaking on first time resize

nvasileiadis:
The more button doesn't ways work when clicked after a window resize. Also, it sometimes shows when it's not needed.


jamesjacobs:
I've pushed fixes for 2 issues:

1. The click event was bound to the more button before it had been added to the dom - resulting in it only loading when the window was loaded with a short enough height.

2. At certain VP heights, after resizing the window the more button would show even when it wasn't needed.

There is still a small issue around when the more button is shown as sometimes when resizing the window, it replaces the last item even when there is enough room for it to fit. I've tried to explore a fix but I've run out of time. As it doesn't really impact the experience I'll look at this in a separate ticket in the future.

.gitignore
```diff
@@ -38,3 +38,4 @@ pulsar.zip
 /stylesheets/pulsar-ie9-theme-*
 /stylesheets/pulsar-theme-*
 /docs/assets/bundle.js
+/.vscode
```
mikedevelops: check you out using Visual Studio, it's pretty cool isn't it 😄

tests/js/web/NavMainComponentTest.js
```diff
@@ -272,18 +272,18 @@ describe('NavMainComponent', function () {
...
+        });
 
-            it('should open the sliding main nav if it is closed', function () {
-                expect(this.$navTertiary.hasClass('is-open')).to.be.true;
-            });
+        it('should open the sliding main nav if it is closed', function () {
+            expect(this.$navTertiary.hasClass('is-open')).to.be.true;
         });
-    });
+    });*/
```
mikedevelops: I know this is discussed in channel, let's fix this or remove. Also whilst on the subject of tests, can we add some to catch the issue this ticket it fixing and the additional code we're adding (unless it is being covered elsewhere)?

PR #777, opened 2018-03-09
Changed the way normalize.css is loaded into pulsar to make it more compatible with sass

skmetaly:
Starting with libsass 3.2 in order to import/include raw CSS-files we need to use the syntax without.css
This results in the actual read of partials s[ac]ss|css and include of it inline to SCSS/SASS


Stanton:
From what I can gather, XFP and CMS are using at least libsass 3.5, so this change _shouldn't_ cause a problem with builds. @jamesjacobs can you verify CXM?

Stanton:

| ---- | ---- |
| require changes to `pulsar.scss` | ⚠️ **Yes, but not explicitly required so not breaking** |

Stanton:
@skmetaly for future PRs, you don't need to namespace them as `feature/` :)

PR #784, opened 2018-03-19
Moved navigation color definitions in theme

skmetaly:
The navigation colors are defined directly in the component. I think they should be defined in `config.theme` so they can be more easily overridden.

**This is not an exhaustive list of what constitutes a breaking change**

| ---- | ---- |
| require changes to `pulsar.scss` | No |
| require changes to `base.html.twig` (or other core views like header/footer) | No |
| require markup changes to maintain functionality | No |

## Do Continuum products use `_config.theme.scss`?

- [x] CMS
- [x] CXM

### DIFF
--- a/stylesheets/_component.navigation.scss
+++ b/stylesheets/_component.navigation.scss
@@ -1,7 +1,3 @@
-$nav-color-light: color(jadu-blue, dark) !default;
-$nav-color-dark: darken(color(jadu-blue, dark), 5%) !default;
-$nav-color-darker: darken(color(jadu-blue, dark), 10%) !default;
-
 .nav-main {
     -webkit-backface-visibility: hidden;
     -webkit-transform: translate3d(-100%, 0, 0);
--- a/stylesheets/_config.theme.scss
+++ b/stylesheets/_config.theme.scss
@@ -126,7 +126,7 @@ $state-info-bg:                  #d9edf7 !default;
 $state-info-border:              darken($state-info-bg, 10%) !default;
 
 // -----------------------------------------------------------------------------
-//  Forms
+//  Navigation
 // -----------------------------------------------------------------------------
 
 $nav-color-light: color(jadu-blue, dark) !default;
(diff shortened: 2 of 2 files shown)


Stanton:
@jamesjacobs Does this interfere with how CXM uses Teal for the primary nav?

Stanton:
@skmetaly This is a sensible change overall, however I think you may be better using the theme classes to override the default colour with your own.

The navigation markup has a selection of `t-` namespaced classes used for theming, these can be targeted by your own styles and might be easier to manage than having to recompile the Sass to deliver a custom theme.

e.g. https://github.com/jadu/pulsar/blob/develop/views/pulsar/v2/helpers/nav.html.twig#L7 (plus more for secondary, tertiary)

We also have a method to automatically generate different themes based on custom palettes, currently this is used to generate projector-friendly versions, but could be used to generate themes if you need this level of customisation. See https://jadu.github.io/pulsar/layout/themes/.

Overall, I think simply targeting the theme classes separately would be the best for you, this is the way CXM allows end-users to customise their own branding (by writing custom CSS into the head, IIRC).

skmetaly:
thanks @Stanton !

I saw how CXM overrides the navigation colours based on each clients settings but i would avoid adding separate styling for the navigation if i can avoid it.

What i ended up doing at least temporarily is creating a new `_palette.deployer.scss` that simply adds a new `jadu-deployer` pallete with the colours i want.

Also i created a new config.theme override that overrides the navigations variables:
```
// -----------------------------------------------------------------------------
//  Navigation
// -----------------------------------------------------------------------------

$nav-color-light: color(jadu-deployer, dark) !default;
$nav-color-dark: darken(color(jadu-deployer, dark), 5%) !default;
$nav-color-darker: darken(color(jadu-deployer, dark), 10%) !default;
```
and removed the variables from _component.navigation.scss by overriding it locally. 

I wanted to take this approach since maybe in the future i would want to override some jadu-blue colors if needed (since i saw there are a number of components similar to navigation one that define the colours directly in the component) 

Thanks again !

Stanton:
@jamesjacobs can you confirm if CXM uses `_config.theme.scss`? If it does this _shouldn't_ be a breaking change...

jamesjacobs:
@Stanton CXM does indeed use `_config.theme`

PR #785, opened 2018-03-22
Accessibility improvements

Stanton:
# Improvements to overall Design System

| Component | Issue | Resolved by |
| ---- | ---- | ---- |
| Sign-in panel | **1.4.3 Contrast (Minimum)** - The color of the text and the color of the background are not in sufficient contrast to each other | Darkened the blue of the using the darkest shade of 'jadu-blue' which meets the 4.5:1 contrast ratio vs. the white link text when an average measurement is taken on screen to factor in the 0.9 opacity |
| Primary navigation links | **1.4.3 Contrast (Minimum)** - The color of the text and the color of the background are not in sufficient contrast to each other | Darkened the blue of the primary navigation by 2.5%, which is the minimum amount to achieve the required 4.5:1 contrast ratio vs. the white link text |
| Back to top link in footer | **1.4.1 Use of Color** - It appears that color is the only distinguishing feature about links in blocks of text | Increased contrast of back to top icon to distinguish from other footer text |

## Lexicon-only improvements

* Added alt text to placeholder images in Cards examples

## Results

Sign in panel

screen shot 2018-03-28 at 15 06 29

Navigation

screen shot 2018-03-28 at 15 17 30


Stanton:

| ---- | ---- |

docs/_styleguides/forms.md
```diff
@@ -5,49 +5,53 @@ title: Forms Styleguide
...
+### Hidden label
+
+Form helpers can use hidden labels which will be ready by screenreaders but not visible in the UI. For helpers, use the `'show-label': false` option to visually hide them.
+
+{% code_example styleguides/hidden-label %}
+
+### aria-labelledby
 
-There may be times when a design calls for an input without an explicit label next to it, but can still be labelled by another element, such as a table heading column label.
+If another visible element on the page can be used to label the input, like a
```
jamesjacobs: like a .....

docs/assets/code_examples/styleguides/twig/aria-label
```diff
@@ -0,0 +1,6 @@
+{{
+    form.checkbox({
+        'id': 'foo',
+        'aria-labelled': 'My awesome checkbox'
```
jamesjacobs: `label`, rather than `labelled`

js/FilterBarComponent.js
```diff
@@ -60,7 +60,7 @@ function createFilterListButton ($filterbar) {
     });
 
     // Build up list and mark up for Add filter button
-    $addFilterButton = $('<button class="btn filter-bar__add" data-ui="show-filter-list" data-toggle="popover" data-trigger="click" title="Filter by" data-html="true" data-placement="bottom" data-content=""><i class="icon-plus"></i></button>');
+    $addFilterButton = $('<button class="btn filter-bar__add" data-ui="show-filter-list" data-toggle="popover" data-trigger="click" title="Filter by" data-html="true" data-placement="bottom" data-content=""><i class="icon-plus"><span class="hide">Add</span></i></button>');
```
jamesjacobs: We should really add `aria-haspopup="true"`,  `aria-expanded="false"` (change to true when clicked and back to false when closed) and  `aria-controls="<<id-of-filter-bar>>"` to this, either in this ticket or a separate one.

docs/_styleguides/forms.md
Stanton: bowss

docs/assets/code_examples/styleguides/twig/aria-label
jamesjacobs: or do you mean `labelledby`?

Stanton:
@jamesjacobs Changes addressed, please re-review. Bo selecta.

PR #791, opened 2018-04-03
Focus first input field after opening a modal

### LINKED ISSUE #754: Opening a modal should focus the first field
When opening a modal, the keyboard tab index stays within the underlying form, the modal component should capture the keyboard tabindex so that the form is immediately useable via keyboard.

Stanton:
This change allows the modal to capture keyboard focus and permit keyboard navigation through embedded forms.

> When a dialog opens, focus moves to an element inside the dialog.
>
> _WAI-ARIA Authoring Practices / 3.8 Dialog (Modal) / Keyboard Interaction <sup>[[1](https://www.w3.org/TR/2017/NOTE-wai-aria-practices-1.1-20171214/#dialog_modal)]</sup>_

### Current behaviour

* Fields are not focused and cannot be navigated to by keyboard alone.
* `autofocus` HTML attribute has no effect in Bootstrap modals and cannot be used <sup>[[2](https://getbootstrap.com/docs/4.0/components/modal/#how-it-works)]</sup>

modal-before

### After

```
Given I have a modal with an input field that can be focused (except buttons)
When I open the modal
Then the field should be focused
```

modal-after

I've started a new test file for any functionality we've added to modals above and beyond the default Bootstrap functionality.

Closes #754 

----
#### References
1. https://www.w3.org/TR/2017/NOTE-wai-aria-practices-1.1-20171214/#dialog_modal
2. https://getbootstrap.com/docs/4.0/components/modal/#how-it-works

### DIFF
--- a/js/libs/modal.js
+++ b/js/libs/modal.js
@@ -248,9 +248,12 @@ var $ = require('jquery');
   $(document)
     .on('show.bs.modal',  '.modal', function () { $(document.body).addClass('modal-open') })
     .on('hidden.bs.modal', '.modal', function () {
-      // $(this).removeData('bs.modal').empty()
       $(document.body).removeClass('modal-open');
+    });
 
-    })
-
+  // Focus the first input in a modal when the modal has been opened
+  $(document).on('shown.bs.modal', '.modal', function() {
+    $(':input:not(input[type=button], input[type=submit], button):visible:first', $(this)).focus();
+  });
+  
 module.exports = Modal;
--- a/tests/js/web/ModalTest.js
+++ b/tests/js/web/ModalTest.js
@@ -11,22 +11,22 @@ describe('Test extra functionality added to Bootstrap modals', function() {
     beforeEach(function() {
         this.$html = $('<div id="html"></div>').appendTo('html');
         this.$body = $('<div id="body"></div>').appendTo(this.$html);
-        this.$code = $('\
-            <a href="#theModal">\
-            <div class="modal fade" id="theModal" tabindex="-1" role="dialog" aria-hidden="true">\
-                <div class="modal__dialog">\
-                    <div class="modal__content">\
-                        <div class="modal__header">\
-                            <button type="button" class="close" data-dismiss="modal" aria-hidden="true">&times;</button>\
-                            <h4 class="modal__title">Standard modal</h4>\
-                        </div>\
-                        <div class="modal__body">\
-                            <input type="text" id="textField" />\
-                        </div>\
-                    </div>\
-                </div>\
-            </div>\
-').appendTo(this.$body);
+        this.$code = $(`
+            <a href="#theModal">
+            <div class="modal fade" id="theModal" tabindex="-1" role="dialog" aria-hidden="true">
+                <div class="modal__dialog">
+                    <div class="modal__content">
+                        <div class="modal__header">
+                            <button type="button" class="close" data-dismiss="modal" aria-hidden="true">&times;</button>
+                            <h4 class="modal__title">Standard modal</h4>
+                        </div>
+                        <div class="modal__body">
+                            <input type="text" id="textField" />
+                        </div>
+                    </div>
+                </div>
+            </div>
+        `).appendTo(this.$body);
 
         this.$modalToggle = this.$html.find('a[href="#theModal"]');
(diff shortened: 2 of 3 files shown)


Stanton:

| ---- | ---- |

tests/js/web/ModalTest.js
```diff
@@ -0,0 +1,58 @@
...
+var $ = require('jquery');
+
+require('../../../js/libs/modal');
+
+describe('Test extra functionality added to Bootstrap modals', function() {
+
+    beforeEach(function() {
+        this.$html = $('<div id="html"></div>').appendTo('html');
+        this.$body = $('<div id="body"></div>').appendTo(this.$html);
+        this.$code = $('\
```
jamesjacobs: No need for these nasty `\` now with ES6 string template literals (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals).

```
$(`<div>
    <lots>
        <of>
             <dom></dom>
        </of>
    </lots>
</div>`);
```

Stanton: Implemented

js/libs/modal.js
```diff
@@ -248,9 +248,12 @@ var $ = require('jquery');
...
     .on('hidden.bs.modal', '.modal', function () {
-      // $(this).removeData('bs.modal').empty()
       $(document.body).removeClass('modal-open');
+    });
 
-    })
-
+  // Focus the first input in a modal when the modal has been opened
+  $(document).on('shown.bs.modal', '.modal', function() {
+    $(':input:not(input[type=button], input[type=submit], button):visible:first', $(this)).focus();
```
mikedevelops: This selector is on the large side, if we are just trying to focus the first focusable input would jQuery's `:focusable` help shorten this at all?

Why are we passing `$(this)` as the context argument to the selector? Shouldn't the context already be set to `.modal` (I could be wrong here, not been in jQuery land for a while), if it's the case that context does need to be explicitly set, we could probably omit creating a jQuery object out of it and just pass in `this`? This is admittedly a really minor probablly-not-worth-doing change!

tests/js/web/ModalTest.js
```diff
@@ -0,0 +1,58 @@
...
+        beforeEach(function() {
+            this.$modalToggle.click();
+            this.$modal.trigger('shown.bs.modal');
+        });
+
+        it('should focus the first input if one is available', function(done) {
+            setTimeout(() => {
+                expect(this.$modalField.is(':focus')).to.be.true;
+                done();
+            }, 500);
```
mikedevelops: instead of this `setTimeout` could we set this up in a way that we don't wait an arbitrary period of time, something like...

```
beforeEach(function() {
    this.$modalToggle.click();
});

it('should focus the first input if one is available', function(done) {
    this.$modal.on('shown.bs.modal', () => {
        expect(this.$modalField.is(':focus')).to.be.true;
        done();
    });

    this.$modal.trigger('shown.bs.modal');
});
```

I know that puts the assertion above the setup code, but I still think that's preferable to a timeout in this instance.

js/libs/modal.js
Stanton: > if we are just trying to focus the first focusable input would jQuery's :focusable help shorten this at all?

We need the first focusable element that's not a button, so wouldn't that just be changing 

`:input:not(input[type=button], input[type=submit], button)` 

to 

`:focusable:not(input[type=button], input[type=submit], button)`

which is longer?

tests/js/web/ModalTest.js
Stanton: aye maybe, one for the next time we touch this file I guess, seeing as it's already merged an ting

js/libs/modal.js
mikedevelops: Right gotcha! Yep good point. I find these pseudo CSS query string really difficult to reason about at a glance, I think I personally prefer chaining jQuery's filter methods, something like 

```$(':input:visible').not('[type="button"],[type="submit"],button).first();```

That is purely my preference though, I think there's some non realisable optimisation as the query scope is reduced in each chain rather than one mega query.

tests/js/web/ModalTest.js
mikedevelops: ah! Haha I'm really sorry, I didn't realised this was merged! Ignore everything.

PR #792, opened 2018-04-03
Addition of Stepped Progressbar

nvasileiadis:
stepped-progressbar-overlap

**This is not an exhaustive list of what constitutes a breaking change**

| ---- | ---- |
| require changes to `pulsar.scss` | No  |
| require changes to `base.html.twig` (or other core views like header/footer) | No |
| require markup changes to maintain functionality | No |


stylesheets/_component.progress-bars.scss
```diff
@@ -1,41 +1,115 @@
...
+        border-radius: 50%;
+        display: block;
+        font-size: 16px;
+        height: 30px;
+        position: relative;
+        text-align: center;
+        width: 30px;
+        zoom: 1;
+
+        &.icon,
```
jamesjacobs: I don't see an example of a `.step-box.icon` in the markup example. Is this line needed?

views/lexicon/tabs/progress_bars.html.twig
```diff
@@ -216,5 +216,32 @@
         </a>
 
     </div>
+    
+    <h2 class="heading">Stepped Progress Bar</h2>
+
+    <div class="stepped-progressbar">
+        <div class="step"><span class="step-box current"><span class="step-number">1</span></span></div>
```
jamesjacobs: I still think this should be an `ol` rather than a collection of `divs` so that screen readers will announce it as a list with x number of items. Adding `role="progressbar"` with `aria-valuenow="1"`, `aria-valuemin="1"` & `aria-valuemax="4"` will also allow screenreaders and assistive technology to recognize it as a progress indicator and announce the current status, for example:

> Progress bar 1.

To make it sound a little nicer, you could add a visually hidden heading, `Step 1 of 4`, give it an `id` and add `aria-describedby="<heading-id>"`, which would give you:

> Progress bar, step 1 of 4, 1.

If we don't want screen readers to announce this at all (I'm not sure why we wouldn't), then we should add `aria-hidden="true"`

stylesheets/_component.progress-bars.scss
Stanton: Should really use `class="icon icon-flag-checkered"` in the examples, the `[class*="icon-"]` is a fallback selector for when the `.icon` parent class isn't present, it's not as performant I don't think...

jamesjacobs: @nvasileiadis Where you have

```
.step-box {
    ... some styles ...
    &.icon,
    [class*='icon-'] {
        ... some styles...
    }
```

It doesn't look like you need the `&.icon`, as in your examples I don't see an element with both the `step-box` *and* the `icon` class. By using `&` it will compile to 

```
.step-box.icon,
.step-box [class*='icon-'] {
```

And it doesn't look like you need the first line, unless I'm missing something?

docs/_components/progress-bar.md
```diff
@@ -105,3 +105,60 @@ If you have the percentage in the label either hidden or visible) then you'll ne
     </div>
 </div>
 ```
+
+## Stepped Progress Bar Example
+
+Stepped progress bar is used mainly in modals as a step indicator of a multi step process.
+<div class="pulsar-example">
```
jamesjacobs: Docs need updating with newer mark up.

nvasileiadis: Updated

PR #799, opened 2018-04-06
Add ability to add styled screenreader samples to documentation

Stanton:
To document the expected behaviour of screenreaders interacting with Pulsar components you can now add the following block as the last child of a `pulsar-example`, use `<var>` to denote user generated content like labels, values or help text.

Samples should be taken with the NVDA screenreader for consistency, documentation will be added to explain that this is to be taken as an approximation. Screenreader samples should be typed phonetically when required, so a screenreader announcing "DD/MM/YYYY" would be typed as "dee-dee slash em-em slash ee-ayy" to more accurately explain how the vocalisation would occur.

```
<p class="screenreader">
    <q><var>First name</var> edit, blank</q>
 </p>
```

e.g.

```
<div class="pulsar-example form">
    <div class="form__group">
        <label for="inputTextSimple" class="control__label">First name</label>
        <div class="controls">
            <input id="inputTextSimple" name="inputText" type="text" class="form__control">
        </div>
    </div>

    <p class="screenreader">
        <q><var>First name</var> edit, blank</q>
    </p>
</div>
```
screen shot 2018-04-06 at 12 01 56


Stanton:
@jamesjacobs Yeah, this PR pre-dates the changes you made to the helpers so the markup and screenreader text will need updating

PR #800, opened 2018-04-09
Fix form helpers/macros to use aria-describedby when help text is used

### LINKED ISSUE #790: Form field help blocks are not read by screenreaders
Currently, when a form field with a help block gains focus, the help block is not read by a screen reader.

Suggested fix: The help block needs a unique ID and the field needs `aria-describedby="<help-block-id>"`.

jamesjacobs:
Previously, if a help text was passed to a form helper/macro it didn't get read by screenreaders. 

Now, if help text is passed to a form helper/macro via the `help` option, a random `guid` ID is created for the help block which the input (or `fieldset` / `form_group` wrapper - depending on the helper type) uses for aria-describedby.

Screen readers now read any supplied help text correctly.

*Note*: this does not cover validation errors - that will be resolved in a separate ticket

Fixed #790

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -78,7 +78,8 @@ module.exports = function(grunt) {
             options: {
                 bin: 'vendor/bin/phpunit',
                 bootstrap: 'tests/unit/bootstrap.php',
-                colors: true
+                colors: true,
+                bin: "php -d memory_limit=-1 ./vendor/bin/phpunit"
             }
         },
 
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -76,10 +76,9 @@ module.exports = function(grunt) {
                 dir: 'tests/unit/'
             },
             options: {
-                bin: 'vendor/bin/phpunit',
+                bin: 'php -d memory_limit=-1 ./vendor/bin/phpunit',
                 bootstrap: 'tests/unit/bootstrap.php',
-                colors: true,
-                bin: "php -d memory_limit=-1 ./vendor/bin/phpunit"
+                colors: true
             }
         },
 
--- a/tests/unit/Jadu/Pulsar/Twig/Macro/Fixtures/form/date-help.html.twig
+++ b/tests/unit/Jadu/Pulsar/Twig/Macro/Fixtures/form/date-help.html.twig
@@ -1,0 +1,8 @@
+{% extends '@cssTests/visual-regression-layout.html.twig' %}
+{% block body %}
+{{
+    form.date(
+    	{'help': 'my help text'}
+    )
+}}
+{% endblock %}
--- a/tests/unit/Jadu/Pulsar/Twig/Macro/Fixtures/form/date-help.html.twig
+++ b/tests/unit/Jadu/Pulsar/Twig/Macro/Fixtures/form/date-help.html.twig
@@ -1,8 +1,6 @@
 {% extends '@cssTests/visual-regression-layout.html.twig' %}
 {% block body %}
 {{
-    form.date(
-    	{'help': 'my help text'}
-    )
+    form.date({'help': 'my help text'})
 }}
 {% endblock %}
--- a/views/pulsar/v2/helpers/elem.html.twig
+++ b/views/pulsar/v2/helpers/elem.html.twig
@@ -43,7 +43,11 @@ value       | string | Specifies the value of the input
         {% set options = options|defaults({'class': 'form__control'}) %}
     {% endif %}
 
-    <input{{ attributes(options|exclude('append bare input_placement prepend naked')) }} />
+    {% if options.help_id is defined and options.help_id is not empty %}
+        {% set options = options|defaults({'aria-describedby': options.help_id}) %}
+    {% endif %}
+
+    <input{{ attributes(options|exclude('append bare input_placement prepend naked help_id')) }} />
 
 {% endspaceless %}
 {% endmacro %}
@@ -182,9 +186,14 @@ value       | string | Specifies the value of the input
 {% spaceless %}
     {% import '@pulsar/pulsar/v2/helpers/util.html.twig' as util %}
 
+
+    {% if options.help_id is defined and options.help_id is not empty %}
+        {% set options = options|defaults({'aria-describedby': options.help_id}) %}
+    {% endif %}
+
     <select{{
         attributes(options
-            |exclude('append bare label prepend selected value options')
+            |exclude('append bare label prepend selected value options help_id')
             |defaults({
                 'class': 'form__control'
             })
@@ -327,9 +336,13 @@ value       | string  | Specifies the value of the input
 {% spaceless %}
     {% import '@pulsar/pulsar/v2/helpers/util.html.twig' as util %}
 
+    {% if options.help_id is defined and options.help_id is not empty %}
+        {% set options = options|defaults({'aria-describedby': options.help_id}) %}
+    {% endif %}
+
     <textarea{{
         attributes(options
-            |exclude('append bare prepend value')
+            |exclude('append bare prepend value help_id')
             |defaults({
                 'class': 'form__control'
             })
--- a/views/pulsar/v2/helpers/elem.html.twig
+++ b/views/pulsar/v2/helpers/elem.html.twig
@@ -47,7 +47,7 @@ value       | string | Specifies the value of the input
         {% set options = options|defaults({'aria-describedby': options.help_id}) %}
     {% endif %}
 
-    <input{{ attributes(options|exclude('append bare input_placement prepend naked help_id')) }} />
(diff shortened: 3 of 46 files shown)


Gruntfile.js
```diff
@@ -76,7 +76,7 @@ module.exports = function(grunt) {
                 dir: 'tests/unit/'
             },
             options: {
-                bin: 'vendor/bin/phpunit',
+                bin: 'php -d memory_limit=-1 ./vendor/bin/phpunit',
```
mikedevelops: Living live on the edge with `-1` 😄

views/pulsar/v2/helpers/form.html.twig
```diff
@@ -116,10 +132,14 @@ data-*        | string | Data attributes, eg: `'data-foo': 'bar'`
     {% import _self as form %}
     {% import '@pulsar/pulsar/v2/helpers/elem.html.twig' as elem %}
 
+    {% if options.help is defined and options.help is not empty %}
+        {% set options = options|merge({ 'help_id': 'guid-' ~ random() }) %}
```
mikedevelops: Super minor, but looks as though the Pulsar convention for strings with spaces is `kebab-case` rather than `snake_case` for macro argument keys, is it worth being consistent?

tests/unit/Jadu/Pulsar/Twig/Macro/Fixtures/form/date-help.html.twig
```diff
@@ -0,0 +1,8 @@
+{% extends '@cssTests/visual-regression-layout.html.twig' %}
+{% block body %}
+{{
+    form.date(
+    	{'help': 'my help text'}
```
Stanton: minor formatting issue with `{}` being on different lines to all other examples

views/pulsar/v2/helpers/elem.html.twig
```diff
@@ -43,7 +43,11 @@ value       | string | Specifies the value of the input
         {% set options = options|defaults({'class': 'form__control'}) %}
     {% endif %}
 
-    <input{{ attributes(options|exclude('append bare input_placement prepend naked')) }} />
+    {% if options.help_id is defined and options.help_id is not empty %}
+        {% set options = options|defaults({'aria-describedby': options.help_id}) %}
+    {% endif %}
+
+    <input{{ attributes(options|exclude('append bare input_placement prepend naked help_id')) }} />
```
Stanton: Shuffle `help_id` to maintain alpha ordering

views/pulsar/v2/helpers/elem.html.twig
```diff
@@ -182,9 +186,14 @@ value       | string | Specifies the value of the input
...
 
+
+    {% if options.help_id is defined and options.help_id is not empty %}
+        {% set options = options|defaults({'aria-describedby': options.help_id}) %}
+    {% endif %}
+
     <select{{
         attributes(options
-            |exclude('append bare label prepend selected value options')
+            |exclude('append bare label prepend selected value options help_id')
```
Stanton: alpha ordering

views/pulsar/v2/helpers/elem.html.twig
```diff
@@ -327,9 +336,13 @@ value       | string  | Specifies the value of the input
...
     {% import '@pulsar/pulsar/v2/helpers/util.html.twig' as util %}
 
+    {% if options.help_id is defined and options.help_id is not empty %}
+        {% set options = options|defaults({'aria-describedby': options.help_id}) %}
+    {% endif %}
+
     <textarea{{
         attributes(options
-            |exclude('append bare prepend value')
+            |exclude('append bare prepend value help_id')
```
Stanton: alpha ordering

views/pulsar/v2/helpers/form.html.twig
```diff
@@ -54,15 +54,31 @@ items         | array  | An array of items to put in the dropdown list (usually
...
+        {{
+            form.group({
+                'parent': options
+                    |only('bare class error guidance guidance-container help help_id id label required show-label type')
+                    |merge({'class': options.class|default ~ ' form__button-group', 'aria-describedby': help_id}),
+                'controls': {'class': 'btn__group'},
+                'inputs': options.items
+            })
+        }}
+    {% else %}
```
Stanton: Can we avoid the duplication here? 

1. set a null help_id at the top of the helper which can be overridden in the loop. e.g. https://twigfiddle.com/aiz1nw
2. move the setting of `aria-describedby` so you don't need a separate form.group instance

```
{% set help_id = null %}
{% if options.help is defined and options.help is not empty %}
    {% set help_id = 'guid-' ~ random() %}
    {% 
        set options = options|merge({ 
            'help_id': help_id,
            'aria-describedBy': 'help_id'
        }) 
    %}
{% endif %}

{{ form.group(all the other stuff) }}
```

views/pulsar/v2/helpers/form.html.twig
Stanton: @mikedevelops there's two things in play here, which I think is what caused the inconsistency, but I don't think this was a conscious decision, we just need to decide whether to keep it or change it:

* If an option ends up in the view (so is really a HTML attribute) it's `kebab-case`, like HTML attributes are
* If an option is internal, it's `snake_case`

views/pulsar/v2/helpers/form.html.twig
```diff
@@ -418,15 +441,34 @@ show-label | bool   | Control visibility of the `<label>` element without affect
...
-    {{
-        form.group({
-            'parent': options
-                |only('bare class error guidance guidance-container help id label required show-label')
-                |exclude('multiple')
-                |defaults({'class': 'form-choice'}),
-            'inputs': inputs
-        })
-    }}
+    {% if options.help is defined and options.help is not empty %}
```
Stanton: same question about duplication as above

views/pulsar/v2/helpers/form.html.twig
```diff
@@ -518,6 +564,20 @@ https://jadu.github.io/pulsar/form-helpers/compound
         {% set groupClass = groupClass ~ ' has-error' %}
     {% endif %}
 
+    {% if options.help is defined and options.help is not empty %}
+        {% set help_id = 'guid-' ~ random() %}
+        {% set options = options|merge({ 'help_id': help_id }) %} {# needed? #}
```
Stanton: comment needed? or is that a question for the reviewer?

views/pulsar/v2/helpers/form.html.twig
```diff
@@ -518,6 +564,20 @@ https://jadu.github.io/pulsar/form-helpers/compound
...
+    <fieldset{{
+        attributes(options
+                |only('class id')
+                |defaults({
+                    'class': groupClass,
+                    'aria-describedby': help_id,
+                })
+            )
+    -}}>
+    {% else %}
```
Stanton: duplication avoidance?

views/pulsar/v2/helpers/form.html.twig
```diff
@@ -1282,7 +1358,7 @@ value  | string | The string or html markup to render inside the help block
 {% spaceless %}
 
     {% if options.value is defined and options.value is not empty %}
-        <span class="help-block">
+        <span class="help-block"{% if options.id is defined and options.id is not empty %} id="{{- options.id|raw -}}"{% endif %}>
```
Stanton: Why is `raw` required here? This will be a problem due to opening up a potential XSS vulnerability.

Stanton:
Documentation code examples will need updating

Stanton:

| ---- | ---- |

views/pulsar/v2/helpers/form.html.twig
mikedevelops: What category would `show-label` fall into? Isn't that technically internal as it's used to derive the visually hidden class on labels? It was seeing that, that made me think `snake_case` wasn't a thing

Stanton: Yep, that's inconsistent ;)

Stanton: I think it'd just be far easier if we stuck with one convention, hyphens most likely as I think the append_type is the main one that appeared differently

mikedevelops: I-think-kebab-as-the-convention-would-work

Stanton: So change to `help-id` and we'll change the rest in due course

Stanton:
Guessing you didn't need to test for the null value in the end? guessing that the Aria attribute isn't in the markup if help_id=null otherwise we'd have failed unit tests. (can't test as on mobile)

jamesjacobs:
@Stanton Correct. I thought I would have to too, but it works as expected without the check. All tests pass.

PR #805, opened 2018-04-17
Close popovers when opening the secondary navigation

Stanton:
Fixes an issue where open popovers would appear over the secondary navigation when it was opened.

(Popovers still need to have a higher z-index than nav to stop them being obscured during normal operation)

### Before
pop-after

### After
pop-after2

### DIFF
--- a/tests/js/web/NavMainComponentTest.js
+++ b/tests/js/web/NavMainComponentTest.js
@@ -74,7 +74,9 @@ describe('NavMainComponent', function () {
                    </div>
                </div>
             </nav>
-            <div class="content-main"></div>
+            <div class="content-main">
+                <a href="#" data-toggle="popover" data-content="foo">popover test</a>
+            </div>
         `).appendTo(this.$body);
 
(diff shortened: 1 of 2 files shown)


tests/js/web/NavMainComponentTest.js
```diff
@@ -494,4 +500,21 @@ describe('NavMainComponent', function () {
             expect(this.$navQuaternary.hasClass('is-open')).to.be.false;
         });
     });
+
+    describe('opening the secondary nav when a popover is open', function () {
+        beforeEach(function () {
+            this.navMainComponent.init();
+            this.clickEvent = $.Event('click');
+            this.$linkOne.trigger(this.clickEvent);
+            this.$popoverLink.trigger(this.clickEvent);
```
jamesjacobs: Are we sure this bit works? I've had issues in the past with triggering the same click event on multiple elements in the same beforeEach.

Stanton: Worked for me, and fails when I comment out the related code. So.... yeah?

PR #810, opened 2018-04-20
Restore has-error / has-danger state classes

Stanton:
This restores the expected light red background to elements with the relevant state class.

I can’t find a legitimate reason for these to have been removed, their restoration may cause a few unexpected situations though…

**This change should not affect forms.** Form helpers have their own has-error styling, the top level `has-error`/`has-danger` classes explicitly avoid being applied to `form__group`.

#### Example usage
`<tr class="has-error">...</tr>`
screen shot 2018-04-20 at 15 36 04

## Breaking change quick checklist

| ---- | ---- |
| require changes to `main.js` | ✅ No  |

### DIFF
--- a/stylesheets/_utility.states.scss
+++ b/stylesheets/_utility.states.scss
@@ -34,10 +34,10 @@ $danger-to-text-color: black;
 
 .has-error:not(.form__group),
 .has-danger:not(.form__group) {
-    background-color: lighten(color(danger), 20) !important;
+    background-color: color(danger, light) !important;
 
     &:hover {
-        background-color: lighten(color(danger), 17) !important;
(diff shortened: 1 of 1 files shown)


stylesheets/_utility.states.scss
```diff
@@ -32,6 +32,15 @@ $danger-to-text-color: black;
...
 
+.has-error:not(.form__group),
+.has-danger:not(.form__group) {
+    background-color: lighten(color(danger), 20) !important;
+
+    &:hover {
+        background-color: lighten(color(danger), 17) !important;
+    }
+}
+
```
mikedevelops: Welcome back to the party pal!

PR #818, opened 2018-04-30
Make table detail column scroll when content exceed viewport

### LINKED ISSUE #814: Table detail column container scrolls rather than contents
table-detail

table-detail-2

Stanton:
Change position from abolute to fixed and add overflow behaviour.

Closes #814

### Before
table-detail-before

### After
table-detail-after

## Browser Testing Checklist

- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge
- [ ] IE 11
- [ ] IE 10
- [ ] IE 9
- [ ] IE 8
- [ ] Mobile Safari
- [ ] Samsung Internet

## Breaking change quick checklist

| ---- | ---- |


stylesheets/_utility.utilities.scss
```diff
@@ -333,3 +333,13 @@ div.no-results {
 .u-margin-bottom-none {
     margin-bottom: 0 !important;
 }
+
+%u-margin-bottom-1,
```
nvasileiadis: What if the name was just 'u-margin-bottom' for this one and 'u-margin-bottom-2x' for the next class? With just 1 and 2 added at the end might not be clear what is their purpose.

jamesjacobs: Yeah, I agree with @nvasileiadis I'd rather we had `u-margin-bottom` and `u-margin-bottom--large` for example. That would make it a little more consistent with other utility classes with modifiers.

PR #832, opened 2018-05-22
Stop disabled links in tabbed navigation triggering a contrast warning

Stanton:
Disabled tabs currentlly contain the following a11y issues:
* Automatic addition of `tabindex="-1"` prevents disabled tabs being navigated to, and makes them impossible to discover with keyboard navigation
* No explicit attributes for screenreaders to discern that tab is disabled

This PR implements the following changes:
* Stop removing disabled tab links from the tabindex, allowing them to be navigated to by keyboard (this may be incorrectly viewed by QA teams as a regression)
* Add `aria-disabled` attribute to allow SRs to determine disabled state, given that we can't add `disabled="true"` to links

Tested by adding `'disabled': true` option to a tab link in the Lexicon demonstrates that disabled tab links can now be navigated to by keyboard.
screen shot 2018-05-22 at 15 19 00

### Voiceover
screen shot 2018-05-22 at 15 24 45

### NVDA
screen shot 2018-05-22 at 15 26 39


views/pulsar/v2/helpers/tabs.html.twig
```diff
@@ -29,9 +29,10 @@
             {% set options = ({}) %}
             <li class="nav-inline__item{% if tab_active is defined and tab_active == tab.id %} nav-inline__item--is-active is-active{% endif %}{% if tab.hidden is defined and tab.hidden == true %} hide{% endif %}">
                 {% if tab.disabled is defined and tab.disabled == true %}
-                    {% set href = '#' %}
+                    {% set href = '#' ~ tab.id %}
                     {% set type = 'link' %}
-                    {% set tabindex = '-1' %}
+                    {% set tabindex = '0' %}
```
mikedevelops: is a tabindex required here, wouldn't we arrive at it at the right time without it?

Stanton: Yeah I spotted this after pushing, already sorted :)

views/pulsar/v2/helpers/tabs.html.twig
```diff
@@ -74,8 +73,7 @@
                         'label': tab.label,
                         'href': href,
                         'data-toggle': type,
-                        'disabled': tab.disabled|default(false),
-                        'tabindex': tabindex
+                        'disabled': tab.disabled|default(false)
```
jamesjacobs: Doesn't this potentially give the link an invalid `disabled` attribute?

Stanton: No, this `options` object is passed through the attribute parser which knows we're working with a link here, so uses `aria-disabled`.

PR #834, opened 2018-06-06
Fix a11y issues in navigation helper

Stanton:
This change addresses some accessibility issues raised during accessibility testing.

* Adds an `aria-label="Home"` to the main branding image
* Use `button` instead of `a` if a navigation link is a fragment identifier instead of a URL, this is the most common use case in Continuum products
* If button is used, move the `href` attribute to `data-target`, so not to create invalid HTML
* Add css to make sure button styles don't leak into the nav
* Turn close menu links into buttons, and make sure they're in the recently added tabindexes
* Revert the recent change to aria-labelledby #868 but add the ability for nav items to specify their own aria-label instead

## Browser testing checklist

- [x] Chrome
- [x] Safari
- [x] Firefox
- [x] Edge
- [x] IE 11
- [x] IE 10
- [x] IE 9
- [x] Mobile Safari
- [x] Samsung Internet

### Screenreaders

| Tool | Tested? | Known issues |
| ---- | ---- | ---- |
| JAWS | Yes | None  |
| NVDA | Yes | None  |
| VoiceOver | Yes | None |

### Accessibility testing tools

| Tool | Tested? | Known issues |
| ---- | ---- | ---- |
| aXe | Yes | None |
| SiteImprove | Yes | None |
| Tenon.io | No |  |
| WAVE | Yes | None |
| Lighthouse | No |  |

## Breaking change checklist

| ---- | ---- |
| require markup changes to maintain functionality | ⚠️ No, but markup changes are required to improve accessibility, but existing hardcoded implementations (like CMS) shouldn't break if they don't implement the changes |


js/NavMainComponent.js
```diff
@@ -130,7 +130,14 @@ NavMainComponent.prototype.manageTabIndexes = function () {
...
 NavMainComponent.prototype.openSecondaryNav = function ($linkClicked, event) {
     var component = this,
+        target;
+
+    if ($linkClicked[0].hasAttribute('href')) {
         target = $linkClicked.attr('href');
+    }
+    else if ($linkClicked[0].hasAttribute('data-target')) {
+        target = $linkClicked.attr('data-target');
+    }
```
mikedevelops: Can we add test coverage here?

Stanton: Added

Stanton:
Fixed issue with button links not having active class and resolved the related code coverage

Issue #835, opened 2018-06-07
Improve consistency of disabled options in helpers


Stanton:
### Problem

There are multiple things that need to be applied to elements when disabling them that maintain accessibility, typically a developer implementing a UI may only toggle the `is-disabled` class, or the `disabled` attribute and not think/know about using `aria-disabled` where appropriate.

The addition of `aria-disabled` also tells a11y tools to not test a disabled element for colour contrast.

PR #836, opened 2018-06-07
Improve consistency of disabled attributes on disabled elements

Stanton:
### Problem

There are multiple things that need to be applied to elements when disabling them that maintain accessibility, typically a developer implementing a UI may only toggle the `is-disabled` class, or the `disabled` attribute and not think/know about using `aria-disabled` where appropriate.

The addition of `aria-disabled` also tells a11y tools to not test a disabled element for colour contrast.

### Proposal

- [x] Update the AttributeParser to make sure the following conditions are met:
```
Given the helper is called with `disabled: true` or `class: 'is-disabled'`
* add the `is-disabled` class, if it hasn't already been passed in `class`
* if an element supports `disabled`, use `disabled`
* if an element does not support `disabled` (like links), use `aria-disabled`
```

- [x] Restrict to `button`, `fieldset`, `input`, `optgroup`, `option`, `select`, `textarea`
- [x] Check all helper documentation includes instructions for disabling the item, with `disabled: true` being the preferred method
- [x] Add documentation for best practices when disabling elements via js, and link to all helper docs
- [x] Update unit tests to cover presence of new attributes

### Potential issues

* May break third-party javascript which expect incorrect `disabled` attributes on links
* May break styling if `is-disabled` class is applied to disabled elements that haven't had it in the past
* May break product-level unit tests which test markup output

## Breaking change checklist

| ---- | ---- |

Addresses #835


docs/_components/badge.md
```diff
@@ -43,12 +43,12 @@ Badges within buttons will inherit the parent button styling without the need to
...
 <div class="pulsar-example">
     <p>
-        <span class="badge">1</span>
-        <span class="badge badge--primary">2</span>
-        <span class="badge badge--success">3</span>
-        <span class="badge badge--warning">5</span>
-        <span class="badge badge--danger">8</span>
-        <span class="badge badge--info">13</span>
-        <span class="badge badge--inverse">21</span>
+        <button class="btn">Base <span class="badge">1</span></button>
```
jamesjacobs: Curious as to why you changed these to buttons rather than stand-alone badges?

docs/_form-helpers/choice.md
```diff
@@ -6,19 +6,28 @@ category: Form helpers
...
-
-          <label class="control__label"><input value="italic" name="foo" type="radio" class="form__control radio">Moonlight</label>
-
-          <label class="control__label"><input value="italic" name="foo" type="radio" class="form__control radio">Good times</label>
-
-          <label class="control__label is-selected"><input value="underline" name="foo" type="radio" checked="" class="form__control radio">Boogie</label>
-
+            <label class="control__label">
+                <input value="sun" name="foo" checked type="radio" class="form__control radio">
+                <span class="form-choice__label">Sunshine</span>
```
jamesjacobs: Is this just updating old doc examples?

PR #840, opened 2018-06-12
Set a white background colour on body to prevent Mojave dark mode from leaking

Stanton:
https://twitter.com/stanton/status/1003971450804555776


Stanton:
@jamesjacobs Reporting and sidebar mobile is fine, was a quick fix and couldn't see a downside to just adding it in

PR #843, opened 2018-06-18
Updated table styling to improve 'obviousness'

Stanton:
* Increases contrast of table header to clearly stand out against horizontal tabs, table tools, actions bar etc
* Improves clarity of sorting options in each header cell
* Fixes minor alignment issues with datatable select options
* Adds new row-actions dropdown examples

**Before**
screenshot 2018-06-18 at 13 54 37

**After**
screenshot 2018-06-18 at 13 50 50

**Row actions dropdown**
screenshot 2018-06-18 at 09 53 54


js/PulsarUIComponent.js
```diff
@@ -158,22 +159,62 @@ PulsarUIComponent.prototype.initDataTables = function () {
         $($.fn.dataTable.tables(true)).DataTable().columns.adjust().responsive.recalc();
     });
 
+    // Dropdowns within responsive tables will be hidden by overflows, this method detaches the dropdown menu and reattaches it on the body
+    $(document).on('show.bs.dropdown', '.datatable', function(e) {
+        var dropdown = $(e.target),
+            dropdownMenu = dropdown.find('.dropdown__menu'),
+            dropdownMenuTop = dropdown.offset().top,
+            dropdownMenuLeft = dropdown.offset().left;
```
mikedevelops: Really tiny, but I'm fairly certain `getBoundingClientRect` (which is what jQuery will be using here) triggers layout/reflow, so we should aim to limit it's use as much as possible. Could we instead store the offset in memory and access the properties from that? Something like...


```
dropdownMenuOffset = dropdown.offset(),
dropdownMenuOffsetTop = dropdownMenuOffset.top,
dropdownMenuOffsetLeft = dropdownMenuOffset.top;
```

js/PulsarUIComponent.js
```diff
@@ -158,22 +159,62 @@ PulsarUIComponent.prototype.initDataTables = function () {
...
+
+        dropdownMenu.css({
+            display: 'block',
+            left: dropdownMenuLeft - dropdownMenu.width() + dropdown.width() - 2,
+            top: dropdownMenuTop + dropdown.height() - 1
+        }).addClass('js-row-actions-dropdown');
```
mikedevelops: Same with `width()` here, although these are admittedly micro optimisations I think we should still strive to avoid reflows where we can in a readable friendly fashion. We could perhaps get the width and store it once.

tests/js/web/PulsarUIComponentTest.js
```diff
@@ -3,10 +3,13 @@
 'use strict'
 
 var $ = require('jquery'),
+    dropdown = require('../../../js/libs/dropdown'),
     history = require('../../../libs/history.js/scripts/bundled/html5/jquery.history'),
     tab = require('../../../js/libs/tab'),
     PulsarUIComponent = require('../../../js/PulsarUIComponent');
 
+    $.fx.off = !$.fx.off;
```
mikedevelops: Do we intentionally want to flip this boolean? Or do we always want it to be `true` for the tests?

tests/js/web/PulsarUIComponentTest.js
```diff
@@ -17,14 +20,35 @@ describe('Pulsar UI Component', function() {
...
+                                    <i class="icon-ellipsis-h"><span class="hide">Actions</span></i>\
+                                </button>\
+                                <ul class="dropdown__menu pull-left">\
+                                    <li><a href="#action">action</a></li>\
+                                </ul>\
+                            </div>\
+                        </td>\
+                    </tr>\
+                </tbody>\
+            </table>\
```
mikedevelops: We could use some purdy ES6 template literals here and get rid of your escaped newlines `\`, it'd just be a case of replacing the single quotes for back ticks:

```
$('<div>\
    <p>Hello, World!</p>\
</div>')

$(`
    <div>
        <p>Hello, world!</p>
    </div>
`)
```

tests/js/web/PulsarUIComponentTest.js
```diff
@@ -108,6 +134,58 @@ describe('Pulsar UI Component', function() {
...
+            sinon.spy(this.pulsarUIComponent, 'closeRowActions');
+            this.pulsarUIComponent.init();
+        });
+
+        it('should be opened when clicked', function() {
+            this.$rowActions.click();
+            setTimeout(function(){
+                expect(this.$rowActions.parent().hasClass('open')).to.be.true;
+            }, 1000);
+        });
```
mikedevelops: If we are testing anything async, we need to utilise the `done` callback, otherwise there is a chance your assertions are never ran and reporting a false positive. Anything with a setTimeout needs a `done`:


```
it('should be opened when clicked', function(done) {
    this.$rowActions.click();
    setTimeout(function(){
        expect(this.$rowActions.parent().hasClass('open')).to.be.true;
        done()
    }, 1000);
});
```

mikedevelops: also on the `setTimeout` used here, how brittle will `1000` be? Does we need to compute this from some CSS or JS animation?

tests/js/web/PulsarUIComponentTest.js
Stanton: I was trying to avoid having to setTimeout just to get some tests to pass, not sure this did anything tbh

mikedevelops: It should disable the animations, whether this makes them behave synchronously (I'm not sure, probably not), my point is if we do want to do this, we should probably be explicit and set ` $.fx.off = true`

tests/js/web/PulsarUIComponentTest.js
Stanton: I don't event think there's animations on the dropdown, just a delay with how long the event takes, this always trips me up

stylesheets/_component.datatables.scss
```diff
@@ -79,27 +78,34 @@ a.dt-button {
         &.sorting_asc::after,
         &.sorting_desc::after {
             @include icon-FontAwesome();
-            font-size: 18px;
+            font-size: 16px;
```
nvasileiadis: Isn't $font-size-base a better approach?

stylesheets/_component.datatables.scss
```diff
@@ -112,14 +118,16 @@ a.dt-button {
 }
 
 .table-row-select {
-    color: #666;
+    color: color(grey);
     cursor: pointer;
-    font-size: 14px;
+    font-size: 16px;
```
nvasileiadis: Similar as the comment above. What if we use $font-size-base?

views/lexicon/tables/datatablex.html.twig
```diff
@@ -40,10 +70,15 @@
                 <td>4 GB</td>
                 <td class="centered">2</td>
                 <td>
-                    {{ html.link({ 'label': 'Reboot', 'class': 'table-action' }) }}<!--
-                    -->{{ html.link({ 'label': 'Suspend', 'class': 'table-action' }) }}<!--
-                    -->{{ html.link({ 'label': 'Shutdown', 'class': 'table-action' }) }}<!--
-                    -->{{ html.link({ 'label': 'Console', 'class': 'table-action' }) }}
+                    {{
+                        html.button_dropdown({
+                            'label': html.icon('ellipsis-h', {'label': 'Actions'}),
```
nvasileiadis: Not sure, if alphabetical order is needed here

views/lexicon/tables/datatablex.html.twig
```diff
@@ -69,10 +105,15 @@
                 <td>2 GB</td>
                 <td class="centered">1</td>
                 <td>
-                    {{ html.link({ 'label': 'Reboot', 'class': 'table-action' }) }}<!--
-                    -->{{ html.link({ 'label': 'Suspend', 'class': 'table-action' }) }}<!--
-                    -->{{ html.link({ 'label': 'Shutdown', 'class': 'table-action' }) }}<!--
-                    -->{{ html.link({ 'label': 'Console', 'class': 'table-action' }) }}
+                    {{
+                        html.button_dropdown({
+                            'label': html.icon('ellipsis-h', {'label': 'Actions'}),
```
nvasileiadis: Same as above about alphabetical order

stylesheets/_component.datatables.scss
```diff
@@ -47,59 +47,76 @@
...
         &.sorting_asc::after,
         &.sorting_desc::after {
             @include icon-FontAwesome();
-            font-size: 18px;
+            font-size: 16px;
             left: 0;
             margin-left: 10px;
         }
 
+        &.sorting::after {
```
jamesjacobs: We need a `not(.sorting_disabled)` in there to prevent the icon being added when sorting is disabled

stylesheets/_component.datatables.scss
```diff
@@ -155,7 +174,8 @@ th.sorting_disabled:hover {
 .dataTables_info {
     display: none;
     float: left;
-    margin-top: 11px;
+    line-height: 30px;
+    margin-left: 12px;
```
jamesjacobs: Can we move this onto `dt-buttons` to prevent info being misaligned when no buttons are present.

jamesjacobs: obvs will need to be margin-right instead

package.json
```diff
@@ -105,6 +105,7 @@
   },
   "dependencies": {
     "babel-polyfill": "^6.26.0",
+    "datatables.net-fixedheader": "^3.1.3",
```
jamesjacobs: Can you expand on what this is for?

docs/_components/datatable.md
```diff
@@ -52,22 +52,30 @@ The `DataTable` plugin is called on any table containing the `datatable` class.
...
+                <th>Company</th>
+                <th>Phone</th>
+                <th data-orderable="false">Actions</th>
             </tr>
         </thead>
         <tbody>
             <tr>
                 <td class="table-responsive">
-                    {{ html.link({ 'class': 'table-child-toggle', 'href': '#', 'label': html.icon('plus-sign') }) }}
+                    <a href="#" class="table-child-toggle"><i class="icon icon-plus-sign">Expand</i></a>
```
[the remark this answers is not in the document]
Stanton: docs are out of date, in the lexicon examples they're just icons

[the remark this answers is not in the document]
Stanton: Changed to buttons

js/libs/dropdown.js
```diff
@@ -148,6 +148,6 @@ var $ = require('jquery');
     .on('click.bs.dropdown.data-api', clearMenus)
     .on('click.bs.dropdown.data-api', '.dropdown form', function (e) { e.stopPropagation() })
     .on('click.bs.dropdown.data-api'  , toggle, Dropdown.prototype.toggle)
-    .on('keydown.bs.dropdown.data-api', toggle + ', [role=menu]' , Dropdown.prototype.keydown)
+    .on('keydown.bs.dropdown.data-api', toggle + ', [role=menu]' , Dropdown.prototype.keydown);
```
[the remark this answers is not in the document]
Stanton: This is bootstrap dropdown code, not ours

stylesheets/_component.datatables.scss
Stanton: Good catch, added

package.json
Stanton: Something I played with but couldn't get working. Removed.

views/lexicon/tables/datatablex.html.twig
Stanton: Not really required

views/lexicon/tables/datatablex.html.twig
Stanton: Not really required

Stanton:
All y'all comments addressed @mikedevelops @jamesjacobs, can you redlight/greenlight pls 🙏

PR #845, opened 2018-06-22
Add docker config and update makefile to resolve recent installation issues

Stanton:
Docker can be used instead of vagrant, which is much faster, start the server with `docker-compose up -d`.

Makefile has been updated with some things that were required for me to get a clean working install after reinstalling MacOS.

### DIFF
--- a/makefile
+++ b/makefile
@@ -36,10 +36,6 @@ build:
 	@ -rbenv global 2.3.0
 	@ echo "\n${CHECK} Done"
 
-	# echo "${HR}\nInstalling Ruby...${HR}\n"
-	# -brew install ruby
-	# echo "\n${CHECK} Done"
-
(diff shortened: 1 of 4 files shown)


makefile
```diff
@@ -29,12 +29,19 @@ build:
...
+	@ echo "${HR}\nInstalling Ruby via rbenv...${HR}\n"
+	@ -brew install rbenv
+	@ -rbenv init
+	@ -rbenv install 2.3.0
+	@ -rbenv global 2.3.0
 	@ echo "\n${CHECK} Done"
 
+	# echo "${HR}\nInstalling Ruby...${HR}\n"
+	# -brew install ruby
+	# echo "\n${CHECK} Done"
```
mikedevelops: Shall we remove this commented code before it's merged, we can always get it from git if anything explodes

Issue #809, opened 2018-04-20
Close the table details pane when clicking outside of the details container


Stanton:
> Given I have opened a table detail panel
> When I click on the darkened area of the UI
> Then the detail panel should close

![details](https://user-images.githubusercontent.com/18653/39054939-d02931e2-44aa-11e8-9504-b349f43ae783.gif)

PR #853, opened 2018-07-30
[809] Close the table details panel when clicking outside of the details container

nvasileiadis:
Related ticket: #809 

This PR adds the ability to close the Table Details from clicking on the backdrop overlay.
pulsar-table details close via backdrop

### DIFF
--- a/js/TableDetailComponent.js
+++ b/js/TableDetailComponent.js
@@ -59,8 +59,8 @@ class TableDetailComponent {
 
         //Close with backdrop click
         this.$html.find('.table-detail-backdrop').on('click', (event) => {
+            event.preventDefault();
             if (this.$html.find('.table-detail-backdrop').hasClass('in')) {
-                event.preventDefault();
                 this.closeDetail();
             }
         });
--- a/js/TableDetailComponent.js
+++ b/js/TableDetailComponent.js
@@ -60,8 +60,9 @@ class TableDetailComponent {
         //Close with backdrop click
         this.$html.find('.table-detail-backdrop').on('click', (event) => {
             if (this.$html.find('.table-detail-backdrop').hasClass('in')) {
+                event.preventDefault();
                 this.closeDetail();
-            };
+            }
         });
     }
     /**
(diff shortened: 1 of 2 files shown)


js/TableDetailComponent.js
```diff
@@ -56,8 +56,15 @@ class TableDetailComponent {
...
         });
-    }
 
+        //Close with backdrop click
+        this.$html.find('.table-detail-backdrop').on('click', (event) => {
+            event.preventDefault();
+            if (this.$html.find('.table-detail-backdrop').hasClass('in')) {
+                this.closeDetail();
+            }
+        });
```
mikedevelops: Does the `'.table-detail-backdrop'` ever get removed from the DOM? If not could we create a variable and then reference it inside the event handler rather than perform the additional `find(...)`.

If we do call `remove(...)` on it, could we instead use `detach(...)` (https://api.jquery.com/detach/)

tests/js/web/TableDetailComponentTest.js
```diff
@@ -115,4 +115,24 @@ describe('TableDetailComponent', () => {
...
 	})
+
+	describe('When the backdrop overlay is clicked', () => {
+		it('should remove the "in" class from the backdrop to hide it', () => {
+			tableDetailComponent.init($body);
+
+			$body.find('[data-table-detail-view-detail]').trigger(clickEvent);
+			$body.find('.table-detail-backdrop').trigger(clickEvent);
+
+			expect($body.find('.table-detail-backdrop').hasClass('in')).to.be.false;
```
mikedevelops: Wouldn't this assertion be true if the `in` class was never added as part of the first triggered click?

could we perhaps make this more of an integration test with:

```
$body.find('[data-table-detail-view-detail]').trigger(clickEvent);

expect($body.find('.table-detail-backdrop').hasClass('in')).to.be.true;

$body.find('.table-detail-backdrop').trigger(clickEvent);

expect($body.find('.table-detail-backdrop').hasClass('in')).to.be.false;
```

tests/js/web/TableDetailComponentTest.js
```diff
@@ -115,4 +115,24 @@ describe('TableDetailComponent', () => {
...
+		});
+
+		it('should remove the "table-detail--open" class from the panel to close it', () => {
+			tableDetailComponent.init($body);
+
+			$body.find('[data-table-detail-view-detail]').trigger(clickEvent);
+			$body.find('.table-detail-backdrop').trigger(clickEvent);
+
+			expect($body.find('[data-table-detail-panel]').hasClass('table-detail--open')).to.be.false;
```
mikedevelops: Same as above. We could be confident that both parts of the functionality are working together by asserting that the `table-detail--open` is added when the first click is triggered.

Issue #855, opened 2018-07-31
Addition of Timeline styles from to Pulsar


nvasileiadis:
For the work of the new CMS Workflow Comment System, there was a need for CMS to get some timeline styles that currently exist only in CXM. The related Pulsar/Playground work for the HTML statics brought them in but we need to isolate them and integrate them smoothly into Pulsar.

PR #857, opened 2018-07-31
[855] Addition of Timeline styles from to Pulsar

nvasileiadis:
The timeline styles are required for the new CMS Workflow Comment System. These changes doesn't affect the previous panel work in Pulsar.

Related ticket: #855 

## Breaking change checklist

| ---- | ---- |
| require changes to `pulsar.scss` |  ⚠️ Yes |
| require markup changes to maintain functionality | ✅ No  |


stylesheets/_component.buttons.scss
```diff
@@ -316,3 +316,16 @@ $btn-white:                 'white';
         }
     }
 }
+
+.view-history-link {
```
Stanton: This shouldn't be in the generic button component, we should try achieve this either through utility classes, or as a part of the timeline styles

stylesheets/_component.panels.scss
```diff
@@ -136,3 +136,7 @@
     margin: 10px 0 0;
     text-align: right;
 }
+
+.panel--danger a.view-history-link {
```
Stanton: Too specific, panels should already set link colours, need to investigate why this isn't happening  

https://github.com/jadu/pulsar/blob/be350a2ce8322122c97d38100c81dda7d82c9a4a/stylesheets/_component.panels.scss#L81

stylesheets/_mixin.responsive-utilities.scss
```diff
@@ -91,3 +91,31 @@
...
+    word-break: break-word; // chrome
+    hyphens: auto; // not supported in chrome
+
+    // trigger word breaks and hyphens in FF
+    @-moz-document url-prefix() {
+        word-break: normal;
+    }
+}
+
+@mixin visuallyhidden {
```
Stanton: You've duplicated an existing mixin called `sr-hide`, the timeline code should be updated to use this

stylesheets/_component.timeline.scss
```diff
@@ -0,0 +1,689 @@
...
+            top: 0;
+            width: 4px;
+
+            .lt-ie10 & {
+                top: 58px;
+            }
+        }
+    }
+
+    .autosave {
```
jamesjacobs: auto is specific to CXM (I think) does it need to be added to Pulsar? Are you using it?

stylesheets/_component.timeline.scss
```diff
@@ -0,0 +1,689 @@
...
+    }
+}
+
+.timeline--no-messages::before {
+    @include respond-min($screen-tablet) {
+        content: none;
+    }
+}
+
+.timeline__notice {
```
jamesjacobs: Again, this is specific to CXM timeline messages - are you other products using this?

stylesheets/_component.timeline.scss
```diff
@@ -0,0 +1,689 @@
...
+        }
+
+        &::before {
+            border-color: rgba(194, 225, 245, 0);
+            border-right-color: $timeline-content-border;
+            border-width: 8px;
+            margin-top: -8px;
+        }
+
+        .timeline__item--new & {
```
jamesjacobs: This is specific to CXM ajaxy timeline messages, does it need to be here?

stylesheets/_component.timeline.scss
```diff
@@ -0,0 +1,689 @@
...
+            &::after {
+                -webkit-animation-duration: 2s;
+                animation-duration: 2s;
+                -webkit-animation-name: newContentAfter;
+                animation-name: newContentAfter;
+                -webkit-transform-style: preserve-3d;
+            }
+        }
+
+        .timeline__item--highlighted & {
```
jamesjacobs: This is specific to CXM ajaxy timeline messages, does it need to be here?

stylesheets/_component.timeline.scss
```diff
@@ -0,0 +1,689 @@
...
+    &__thumbnail {
+        max-height: 250px;
+        max-width: 100%;
+
+        @include ie-lte(10) {
+            max-width: 150px;
+        }
+    }
+
+    &__changes {
```
jamesjacobs: This is specific to CXM timeline "diffs" where we show data changing from one value to another, does it need to be in Pulsar? Same with all the nested and diff selectors.

stylesheets/_component.buttons.scss
nvasileiadis: I'm afraid it cannot be utility class because they are a bit inflexible regarding responsiveness. Will move them into timeline styles.

stylesheets/_mixin.responsive-utilities.scss
nvasileiadis: Updated as advised

stylesheets/_component.timeline.scss
```diff
@@ -0,0 +1,585 @@
...
+    }
+
+    &__to {
+        background: #abdfab;
+        border-radius: 3px;
+        color: #3e6441;
+        padding: 1px 2px;
+    }
+
+    &__diff-toggle {
```
jamesjacobs: Unless you're using timeline diffs (which AFAIK) is only a CXM thing, I'd remove all these `diff` selectors

stylesheets/_component.timeline.scss
```diff
@@ -0,0 +1,585 @@
...
+    }
+
+    &__note-title {
+        margin: 1em 0;
+    }
+
+    &__text {
+        @include word-wrap();
+
+        margin: .5em 0;
```
jamesjacobs: selector specific styles should be above includes

stylesheets/_component.timeline.scss
```diff
@@ -0,0 +1,585 @@
...
+    @include sr-hide;
+}
+
+.timeline__form__controls {
+    @include clear-fix(true);
+
+    padding: 0;
+}
+
+.timeline__status-select {
```
jamesjacobs: CXM specific

stylesheets/_component.timeline.scss
```diff
@@ -0,0 +1,585 @@
...
+    margin-top: 10px;
+    width: auto;
+
+    @include respond-min($screen-tablet) {
+        margin-top: 0;
+    }
+}
+
+.timeline__form__control,
+.timeline__status-undo {
```
jamesjacobs: CXM specific

stylesheets/_mixin.responsive-utilities.scss
```diff
@@ -91,3 +91,20 @@
         @include sr-hide;
     }
 }
+
+@mixin word-wrap() {
+    -ms-word-break: break-all; // ie11
+    -ms-hyphens: auto;
```
jamesjacobs: Do we need all these ms specific prefixes? autoprefixer should handle these.

stylesheets/_component.timeline.scss
```diff
@@ -0,0 +1,585 @@
...
+    pointer-events: none;
+
+    &:hover,
+    &:focus,
+    &:active {
+        color: initial;
+    }
+}
+
+.timeline__attach-file-error {
```
jamesjacobs: Needed?

stylesheets/_component.timeline.scss
```diff
@@ -0,0 +1,585 @@
...
+    color: color(danger);
+    font-size: 14px;
+    line-height: 1.5;
+    margin: .675em 0 0;
+}
+
+.timeline__attachment-feedback .timeline__attach-file-error {
```
jamesjacobs: Needed?

stylesheets/_component.timeline.scss
```diff
@@ -0,0 +1,585 @@
...
+
+.timeline__form__control,
+.timeline__status-undo {
+    font-size: 14px;
+    line-height: 26px;
+    margin: 0 1em 0 0;
+    text-decoration: none;
+}
+
+.timeline__undo-link {
```
jamesjacobs: Needed?

stylesheets/_component.timeline.scss
jamesjacobs: L361 to L414

stylesheets/_mixin.responsive-utilities.scss
jamesjacobs: Also, should this be in responsive-utilities as it's not responsive specific? Would be better to follow convention and add a new `_mixin.word-wrap.scss` file.

stylesheets/_component.timeline.scss
```diff
@@ -0,0 +1,585 @@
...
+        padding: 0;
+    }
+}
+
+.timeline__item {
+    margin: 1em 0;
+    position: relative;
+
+    @include respond-min($screen-tablet) {
+        //margin: 2em 0;
```
mikedevelops: We should remove commented code, we can always get it back with git if we need it down the line.

stylesheets/_component.timeline.scss
```diff
@@ -0,0 +1,585 @@
...
+
+    &--system {
+        background-color: color(jadu-blue);
+        color: color(white);
+    }
+}
+
+.timeline__message-box {
+    .timeline-content {
+        background-color: darken(#fafafa, 5%);
```
mikedevelops: Should this use one of the colours from our palette?

stylesheets/_component.timeline.scss
```diff
@@ -0,0 +1,585 @@
...
+
+        @include respond-min($screen-tablet) {
+            &::after {
+                border-color: rgba(136, 183, 213, 0);
+                border-right-color: darken(#fafafa, 5%);
```
mikedevelops: Can we use a palette colour instead? (or add one to the palette)

stylesheets/_component.timeline.scss
```diff
@@ -0,0 +1,585 @@
...
+        display: none;
+
+        @include respond-min($screen-tablet) {
+            display: block;
+        }
+    }
+}
+
+.timeline-content {
+    background: #fafafa;
```
mikedevelops: Can we use a palette colour instead? (or add one to the palette)

stylesheets/_component.timeline.scss
```diff
@@ -0,0 +1,585 @@
...
+            height: 0;
+            position: absolute;
+            right: 100%;
+            top: 16px;
+            width: 0;
+        }
+
+        &::after {
+            border-color: rgba(136, 183, 213, 0);
+            border-right-color: #fafafa;
```
mikedevelops: Can we use a palette colour instead? (or add one to the palette)

stylesheets/_component.timeline.scss
```diff
@@ -0,0 +1,585 @@
...
+            content: '';
+            height: 0;
+            position: absolute;
+            right: 100%;
+            top: 16px;
+            width: 0;
+        }
+
+        &::after {
+            border-color: rgba(136, 183, 213, 0);
```
mikedevelops: Can we use a palette colour instead? (or add one to the palette)

stylesheets/_component.timeline.scss
```diff
@@ -0,0 +1,585 @@
...
+            border-right-color: #fafafa;
+            border-width: 7px;
+            margin-top: -7px;
+        }
+
+        &::before {
+            border-color: rgba(194, 225, 245, 0);
```
mikedevelops: Can we use a palette colour instead? (or add one to the palette)

stylesheets/_component.timeline.scss
```diff
@@ -0,0 +1,585 @@
...
+            border-right-color: $timeline-content-border;
+            border-width: 8px;
+            margin-top: -8px;
+        }
+    }
+
+    &__title {
+        color: #303e49;
```
mikedevelops: Can we use a palette colour instead? (or add one to the palette)

stylesheets/_component.timeline.scss
```diff
@@ -0,0 +1,585 @@
...
+        }
+    }
+
+    &__from {
+        display: inline-block;
+        height: 0;
+        opacity: 0;
+        overflow: hidden;
+        width: 0;
+    }
```
mikedevelops: if the intention here is to visually hide the element, should we use our existing visually-hidden mixin / utility class for this?

stylesheets/_component.timeline.scss
```diff
@@ -0,0 +1,585 @@
...
+            text-decoration: none;
+        }
+    }
+
+    &__to {
+        background: #abdfab;
+        border-radius: 3px;
+        color: #3e6441;
+        padding: 1px 2px;
+    }
```
mikedevelops: lots more hardcoded colours here 👆 , could become a bit of a pain when we need to make sweeping brand or contrast changes

mikedevelops:
A lot of my comments are regarding use of hardcoded colours @nvasileiadis, I stopped commenting on each one towards the end, so if you agree, it's probably worth doing a pass over the file and see if we can convert those hardcoded hex and rgb values to variables.

stylesheets/_component.timeline.scss
nvasileiadis: Scssllint complaints with: "Rule sets should be ordered as follows: `@extends`, `@includes` without `@content`, properties, `@includes` with `@content`, nested rule sets". Therefore should remain as is

stylesheets/_mixin.responsive-utilities.scss
Stanton: We should maybe think about creating a separate include file to include mixins so we don't have to make a breaking change every time we want to introduce a new mixin...

Separate ticket for that though

If this is currently not a breaking I'd leave it in responsive and we can look at introducing a mixin includes file as a separate breaking change

PR #865, opened 2018-08-10
Add tabindexes to toolbar elements and make sure a logical tab order is maintained

Stanton:
This fixes the following issues:

* Make sure toolbar elements are tabbed in the correct order
* Make sure main navigation is tabbed to once the user clicks the mobile menu button

Required to maintain compliance with WCAG 2.4.3 Focus Order: (Level A).

The following components will need a tabindex attribute adding.

| Component | Tabindex |
| ---- | ---- |
| `.skip-link` | 1 |
| `.jadu-branding` | 1 |
| `.nav-link` in primary nav | 1 |
| `.mobile-menu-button` | 2 |
| `.main-search__field` | 3 |
| `.main-search__submit` | 3 |
| `nav.help()` helper (user guides menu) | 4 |
| `.user-menu` | 5 |
| `.notifications-toggle` | 6 |

## Breaking change checklist

| ---- | ---- |
| require changes to `pulsar.scss` | No |
| require changes to `base.html.twig` (or other core views like header/footer) | ⚠️Yes |
| require markup changes to maintain functionality | ⚠️Core elements listed above will need tabindex values adding to the base template |

## Browser testing checklist

- [x] Chrome
- [x] Safari - _requires option-tab or the 'press tab to highlight each item' option checked in preferences > advanced (this is default safari behaviour)_
- [x] Firefox
- [x] Edge
- [x] IE 11
- [x] IE 10
- [x] IE 9

### DIFF
--- a/js/NavMainComponent.js
+++ b/js/NavMainComponent.js
@@ -4,6 +4,7 @@ var $ = require('jquery');
 
 function NavMainComponent ($html, rootWindow) {
     this.$html = $html;
+    this.window = rootWindow;
     this.$window = $(rootWindow);
 };
 
@@ -24,6 +25,7 @@ NavMainComponent.prototype.init = function () {
     component.$body = this.$html.find('body');
     component.$navMain = this.$html.find('.nav-main');
     component.$contentMain = this.$html.find('.content-main');
+    component.$brandingLink = this.$html.find('.jadu-branding');
     component.$navPrimary = this.$html.find('.nav-primary');
     component.$navSecondary = this.$html.find('.nav-secondary');
     component.$navTertiary = component.$navMain.find('.nav-tertiary');
@@ -36,6 +38,9 @@ NavMainComponent.prototype.init = function () {
     // Calculate what primary nav items can be shown and which need to be hidden in more menu
     component.adjustNavItems();
 
+    // Check which tabindexes should be applied to navigation links to ensure WCAG compliance
+    component.manageTabIndexes();
+
     // Open navigation on mobile
     component.$mobileMenuButton.on('click', function(event) {
         var $self = $(this);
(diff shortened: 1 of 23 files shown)


Stanton:
Would say this could go through as a minor, the changes to the base templates aren't breaking nor required to maintain functionality, they're just required to improve accessibility.

js/NavMainComponent.js
```diff
@@ -4,6 +4,7 @@ var $ = require('jquery');
 
 function NavMainComponent ($html, rootWindow) {
     this.$html = $html;
+    this.window = rootWindow;
     this.$window = $(rootWindow);
```
nvasileiadis: Could you please explain why we need both window and $window? What are the limitations of the code. Is it only for the stub in the testing? I don't understand that part

Stanton: `this.$window` already existed in this test but wouldn't work with matchMedia, which required the raw `window`. Looking at it now I can probably just pass `rootWindow` to matchMedia instead so I'll refactor 🤔

Stanton: not going to refactor actually, they're for different things, some functions use the jQuery object, matchMedia needs the raw window object

js/NavMainComponent.js
```diff
@@ -46,9 +51,13 @@ NavMainComponent.prototype.init = function () {
         if ($self.text() === 'Menu') {
             $self.text('Close');
             $self.attr('aria-expanded', 'true');
+            component.$brandingLink.attr('tabindex', '3');
+            component.$primaryNavLinks.attr('tabindex', '3');
```
nvasileiadis: aXe finds them wrong. Elements should not have tabindex greater than zero -https://dequeuniversity.com/rules/axe/1.1/tabindex Is there another reason we use a greater number?

Stanton: aXe is rather strict for this condition, I believe this change to meet the 2.4.3 guideline.

From technique H4: Creating a logical tab order through links, form controls, and objects
> In some cases, the author may want to specify a tab order that follows relationships in the content without following the order of the interactive elements in the code. In these cases, an alternative order can be specified using the tabindex attribute of the interactive element. The tabindex is given a value between 0 and 32767.
https://www.w3.org/TR/WCAG20-TECHS/H4.html

And from F44: Failure of Success Criterion 2.4.3 due to using tabindex to create a tab order that does not preserve meaning and operability
> If tabindex is used, check that the tab order specified by the tabindex attributes follows relationships in the content.
https://www.w3.org/TR/WCAG20-TECHS/F44.html

Stanton: This has also been raised with aXe team: 
https://github.com/dequelabs/axe-core/issues/572

Stanton:
Skip link added as first focusable item

Issue #801, opened 2018-04-12
Date helper does not support international date formats


jamesjacobs:
Current date format is set to `DD/MM/YYYY`. There is also no way to disable pikaday.
The helper should be flexible to accept different date formats for use in the placeholder and pikaday.

PR #873, opened 2018-08-16
[801] Date helper does not support international date formats

nvasileiadis:
PR related to issue #801 

This PR adds the options for the dates to be formatted with US or reverse styles. Based on [Wikipedia](https://en.wikipedia.org/wiki/Date_format_by_country) we are covering all the 3 popular formats that are used around the globe in different combinations. It also handles the case we just need a bare input for the date without any date pickers.

timeformats

## Breaking change checklist

| ---- | ---- |
| require changes to `main.js` | ✅ No  |

## Browser testing checklist

- [x] Chrome
- [x] Safari
- [x] Firefox
- [x] Edge
- [x] IE 11
- [x] IE 10
- [x] IE 9
- [x] IE 8
- [x] Mobile Safari
- [x] Samsung Internet

### Screenreaders

| Tool | Tested? | Known issues |
| ---- | ---- | ---- |
| JAWS | Yes | Will be addressed in #884  |
| NVDA | Yes | Will be addressed in #884 |
| VoiceOver | Yes | Will be addressed in #884 |

### Accessibility testing tools

| Tool | Tested? | Known issues |
| ---- | ---- | ---- |
| aXe | Yes | None |
| SiteImprove | Yes | None |
| Tenon.io | No |  |
| WAVE | Yes | None |
| Lighthouse | Yes | None |


docs/_form-helpers/date.md
```diff
@@ -47,6 +47,32 @@ data-*      | string | Data attributes, eg: `'data-foo': 'bar'`
 
 Any other options not listed here will be applied to the input.
 
+## US and Reverse Formats
+
+Add the `data-datepickerStyle': 'US'` or `'data-datepickerStyle': 'reverse'` option to format the date properly.
```
Stanton: Can we turn this into a normal `format` option for the helper instead of a data attribute? it can turn into a data attribute in the actual code, but helper options are usually simpler

Stanton: Are these the only options we need to support, or should we simply accept a standard date time formatted string?

js/PulsarFormComponent.js
```diff
@@ -85,10 +89,30 @@ PulsarFormComponent.prototype.initTimePickers = function () {
...
     this.$html
-        .find('[data-datepicker=true]')
+        .find('[data-datepickerStyle=default]')
         .pikaday({ format: 'DD/MM/YYYY' });
 }
 
+/**
+ * Initiate a date picker on data-datepicker fields using pickaday US Format
+ */
+PulsarFormComponent.prototype.initUSDatePickers = function () {
```
Stanton: if we accepted a format string we wouldn't need these separate initialisers, we could just pass the string straight to pikaday's format...

docs/_form-helpers/date.md
nvasileiadis: Based on the wikipedia article and our current needs I would suggest we go with those until there is a feature request for something more complex. We are speaking only about dates and not datetime strings. Will simplify the rest based on your first comment

mikedevelops: Sorry to be the guy that says "Shouldn't we name this differently", but ... Shouldn't we name this differently? 

I think `datePickerStyle` could be ambiguous in this context, should we consider calling this `datePickerFormat` (to coincide with the wiki) and to remove the possible assumption that this option is tied to styling (appearance / css)

js/PulsarFormComponent.js
```diff
@@ -84,9 +84,24 @@ PulsarFormComponent.prototype.initTimePickers = function () {
...
 PulsarFormComponent.prototype.initDatePickers = function () {
     // Attach basic pikaday to datepicker fields
-    this.$html
-        .find('[data-datepicker=true]')
-        .pikaday({ format: 'DD/MM/YYYY' });
+
+    const datepickers = this.$html.find('[data-datepicker="true"]');
+
+    datepickers.each((index, element) => {
+        const $element = $(element);
```
mikedevelops: I think we should cache the parent here instead of the element, as it looks as though it's the parent which is used as the base for the `.find(...)` query.

Something like...

```
const $parent = $(element).parent();

$parent.find(...).pikaday();
$parent.find(...).pikaday();
$parent.find(...).pikaday();
$parent.find(...).pikaday();
$parent.find(...).pikaday();
$parent.find(...).pikaday();
```

Also on this, is there a reason we find the datepicker's parent and then go back down into the datepicker. As we're controlling the markup here, can't we just check the value of `data-datepickerStyle` on the `element` in this context?

nvasileiadis: The data-datepickerStyle and data-datepicker are within the same input

Stanton:
If you need pointers on why the travis builds are failing let me know 👍

js/PulsarFormComponent.js
mikedevelops: That is why I ask. In the code for each `data-datepicker` element you find it's parent then query an attribute that exists on the `data-datepicker` element. When we would simple check the value of that attribute on that element directly...

```
things.forEach((index, element) => {
    // ?
    const format = $(element).attr('data-datePickerStyle');
});
```

tests/js/web/PulsarFormComponentTest.js
```diff
@@ -259,7 +259,7 @@ describe('PulsarFormComponent', function() {
         it('should initiate the date pickers', function () {
             this.pulsarForm.refresh();
 
-            expect($.fn.pikaday).to.have.been.calledOnce;
+            expect($.fn.pikaday).to.have.been.called;
```
mikedevelops: Can we remain explicit on the call count here in order to make sure we aren't ever calling it an unexpected amount of times?

With sinon we have `calledOnce`, `calledTwice`, `calledThrice` and `callCount` (that returns the call count as a number)

docs/_form-helpers/date.md
nvasileiadis: Because we are speaking about html attributes, everything that doesn't exist already should be data attribute to future-proof our code. As @mikedevelops mentioned, if in HTML6 for example they decide to add format for other use we will need to change it.

Stanton: For the vast majority of helpers we've gone with simple options, I'd much prefer this to be `format` rather than a data attribute for the helper's options object, please.

docs/_form-helpers/date.md
```diff
@@ -43,10 +43,37 @@ name        | string | The name of this control
 placeholder | string | A short hint that describes the expected value
 required    | bool   | Adds `required` and `aria-required="true"` attributes
 value       | string | Specifies the value of the input
+data-datepicker | bool | Initialises or not the datepicker. Accepts `true` (default) or `false`
```
Stanton: Can you please change this back to `format` so it's consistent with the other options for this helper, the markup it outputs can still be a data attribute

js/PulsarFormComponent.js
```diff
@@ -84,9 +84,24 @@ PulsarFormComponent.prototype.initTimePickers = function () {
...
+            .pikaday({ format: 'DD/MM/YYYY' });
+
+        $parent
+            .find('[data-format=US]')
+            .pikaday({ format: 'MM/DD/YYYY' });
+
+        $parent
+            .find('[data-format=reverse]')
+            .pikaday({ format: 'YYYY/MM/DD' });
+    });
```
mikedevelops: I'm still not certain all of these `find(...)` calls are necessary.

You will be able to get the `data-format` attribute within this `datepickers.each(...)` with `$(element).attr('data-format')`. I'm unclear as to why we need to find the parent and then query it, this approach would make more sense if the `data-format` and `data-datepicker` existed on separate elements.

Stanton: This change will break any existing datepickers which don't have a format option defined, there should be a default fallback to UK dates when no format is supplied.

This should be addressed while refactoring out the multiple find calls.

js/PulsarFormComponent.js
```diff
@@ -84,9 +84,24 @@ PulsarFormComponent.prototype.initTimePickers = function () {
  */
 PulsarFormComponent.prototype.initDatePickers = function () {
     // Attach basic pikaday to datepicker fields
-    this.$html
-        .find('[data-datepicker=true]')
-        .pikaday({ format: 'DD/MM/YYYY' });
+
+    const datepickers = this.$html.find('[data-datepicker="true"]');
+
+    datepickers.each((index, element) => {
```
Stanton: This change will break any existing datepickers which don't have a format option defined, there should be a default fallback to UK dates when no format is supplied.

This should be addressed while refactoring out the multiple find calls.

js/PulsarFormComponent.js
mikedevelops: If you add this to `line 91`

```console.log($(element).attr('data-format'));```

You should see what I'm getting at, you have access to the format without having to query parents and descendants.

Comment (Stanton) [reply js/PulsarFormComponent.js]: @nvasileiadis That unfortunately presumes developers will only use the helper, and doesn't cater for hardcoded markup

File: js/PulsarFormComponent.js (hunk 7 above)
Comment (nvasileiadis) [reply js/PulsarFormComponent.js]: There is 'data-format': options.format|default('default') in the helper. Isn't it enough?

Comment (Stanton) [reply js/PulsarFormComponent.js]: @nvasileiadis That unfortunately presumes developers will only use the helper, and doesn't cater for hardcoded markup

**Comment by nvasileiadis** [comment]
@Stanton I will restore default case in switch because when it is absent I am getting the date format stuck at reverse.

File: js/PulsarFormComponent.js [hunk 8]
```diff
@@ -83,10 +83,31 @@ PulsarFormComponent.prototype.initTimePickers = function () {
...
-        .pikaday({ format: 'DD/MM/YYYY' });
+    const datepickers = this.$html.find('[data-datepicker="true"]');
+    let defaultDateFormat = 'DD/MM/YYYY';
+
+    datepickers.each((index, element) => {
+        const dateFormat = element.getAttribute('data-format');
+
+        // Add missing data-format to hardcoded examples
+        if (!dateFormat) {
+            element.setAttribute('data-format', 'default');
```
mikedevelops: Why do we need to add this attribute? I was under the impression that this component uses it to determine the format, so if the attribute is missing, can't we just use the default format and not add the attribute?

mikedevelops: How does that make this backwards compatible? I can't see anything in this service to suggest that the attribute value of `data-format` is used again.

Issue #878, opened 2018-09-11
CMS wide warning related to PHP version


nvasileiadis:
Every page load within the CMS throws the below

`[06-Sep-2018 12:00:55 Europe/London] PHP Warning:  count(): Parameter must be an array or an object that implements Countable in /var/www/jadu/vendor/jadu/pulsar/src/Jadu/Pulsar/Twig/Extension/HelperOptionsModifierExtension.php on line 59`
Related to the breaking change here - #827
 
Note: The server this is observed on is on PHP 7.2

nvasileiadis:
Suggested fix from @abul-wcse is:

```
<?php

namespace Jadu\Pulsar\Twig\Extension;

class HelperOptionsModifierExtension extends \Twig_Extension
{
    /**
     * Name of this extension
     *
     * @return string
     */
    public function getName()
    {
        return 'modify_options';
    }

    /**
     * Register the `modify_options()` function with Twig.
     *
     * @return array The Twig function
     */
    public function getFunctions()
    {
        return array(
            new \Twig_SimpleFunction('modify_options', array($this, 'modifyOptions'))
        );
    }

    /**
     * Returns a modified options array including error flag, error and help guids
     *
     * @param  array    $options    The form helper options
     * @return array|boolean
     */
    public function modifyOptions($options)
    {
        // If we don't have $options, abandon ship
        if ($options === null) {
            return false;
        }

        // Check if errors are present in options
        if (array_key_exists('error', $options) and is_array($options['error']) and count($options['error']) > 0) {
            $numberOfErrors = count($options['error']);
            $errorGuids = array();

            // Set has_error for aria-invalid use
            $options['has_error'] = true;

            // Build guids for errors
            for ($i = 0; $i < $numberOfErrors; $i++) {
                $errorGuids[] = 'guid-' . rand();
            }

            $options['error_ids'] = $errorGuids;
        }

        // Check if errors are present in options
        if (array_key_exists('help', $options) and is_array($options['help']) and count($options['help']) > 0) {
            $options['help_id'] = 'guid-' . rand();
        }

        return $options;
    }
}
```

Issue #875, opened 2018-09-06
Datatable Separator Issue


nvasileiadis:
Related to [CMS-5604](https://jadultd.atlassian.net/browse/CMS-5604) 

![actions](https://user-images.githubusercontent.com/1425876/45161812-8e213180-b1e4-11e8-8395-7941be1a3e57.png)
![space_separator](https://user-images.githubusercontent.com/1425876/45161814-8eb9c800-b1e4-11e8-98ab-a7941883777b.png)

Getting dotted border-right and dotted border-bottom.

PR #879, opened 2018-09-11
CMS Fixes for Pulsar 8.1.0

nvasileiadis:
Related ticket 
CMS-5602

Issue 1: #878 
Issue 2 and 5: #875

## Breaking change checklist

| ---- | ---- |
| require changes to `gruntfile.js` |✅ No |


stylesheets/_component.tables.scss
```diff
@@ -158,6 +158,7 @@
 }
 
 .table-action {
+    border-bottom: none !important;
```
jamesjacobs: Could we work around using `!important` here? I'm guessing the aim is to remove the bottom border recently added by `.table [href]` - which seems a little too unspecific? Could we address the issue there perhaps? Either change it to use a more specific selector or chain `.table-action` here to bump specificity? Perhaps it's time to remove the right-hand separators on `.table-action`? I'd rather we kept most if not all standard links underlined. Alternatively, you could remove the `table-action` class from the links in CMS (which is what I did in CXM).

stylesheets/_component.datatables.scss
```diff
@@ -373,3 +373,23 @@ th.sorting_disabled:hover {
 }
 
 // scss-lint:enable SelectorFormat
+
+.dtr-details {
+    .dtr-title {
+        a {
```
jamesjacobs: Is this block to address the same issue as below (underlined links?) if so why are the extra `color` and `font-family` styles needed?

Stanton: colour and font family are used to switch styles when DT goes into the (+) responsive mode, because the links move out of the `th` they adopt regular `td` styles when we want them to still look like headers, they also lose their interactive features so if I remember rightly, Nikos and I discussed just resetting the styles like this.

stylesheets/_component.tables.scss
Stanton: > Perhaps it's time to remove the right-hand separators on .table-action

I'd agree with that, the underlining of the links should be enough to visually separate actions now

src/Jadu/Pulsar/Twig/Extension/HelperOptionsModifierExtension.php
```diff
@@ -40,7 +40,7 @@ public function modifyOptions($options)
         }
 
         // Check if errors are present in options
-        if (array_key_exists('error', $options) and count($options['error']) > 0) {
+        if (array_key_exists('error', $options) and is_array($options['error']) and count($options['error']) > 0) {
```
Stanton: I'd swap these if only for readability sake, make it read as the `is_array` is the first thing we check

stylesheets/_component.datatables.scss
jamesjacobs: 👍

src/Jadu/Pulsar/Twig/Extension/HelperOptionsModifierExtension.php
nvasileiadis: Done

PR #882, opened 2018-09-18
Fix invalid markup for required and guidance options in form helpers

### LINKED ISSUE #880: Guidance and required indictor use invalid markup
The rel attribute cannot be used here and results in invalid markup.

**Required example:**
```
<span class="required-indicator" rel="tooltip" data-toggle="tooltips" title="" data-original-title="required">*</span>
```

From https://html5.validator.nu/:
```
Error: Attribute rel not allowed on element span at this point.
```

**Guidance example:**
```
<i data-container="body" data-content="Help text to give more information about expected input" data-placement="top" rel="clickover" aria-hidden="true" class="icon-question-sign input-group-guidance" data-original-title="" title=""></i>
```

From https://html5.validator.nu/:
```
Error: Attribute rel not allowed on element span at this point.
```

----

## Supplemental

- [x] The guidance popover is contained within the `<label>` element and any clicks to that element trigger the input focus

checkbox

Stanton:
This change adds clickover-style options to the default popovers and removes the need to add `rel="clickover"` to elements which shouldn't have the `rel` attribute.

Now, by adding `data-toggle="popover" data-autoclose="true"` this will close popovers when the user clicks anywhere else in the UI.

* Fixes an issue where clicking the guidance icon would toggle the selected state of a checkbox/radio input.
* Removes the `bootstrapx-clickover` library from bower
* Continue support for `rel="clickover"` from within the default popover js

Fixes #880

## Breaking change checklist

| ---- | ---- |
| require changes to `main.js` | 🛑 YES |
| require changes to `index.js` | 🛑 YES |

## Breaking change remediation

1. Remove any reference to `clickover` on your `main.js`, and any other product specific javascript that initialises `clickover` manually.

e.g. remove:
```
// Use clickover enhancements for popovers
$('[rel="clickover"]').clickover({ 'global_close': true });
```

2. Remove any reference to `clickover` from `index.js`, which will now throw an error because the library no longer exists within Pulsar.

e.g, remove:
```
clickover = require('../libs/bootstrapx-clickover/js/bootstrapx-clickover'),
```

3. Check your filesystem for any references to the clickover package, this was previously bundled within the Pulsar `/libs` directory, managed by Bower, and should not (in theory) exist within a Continuum product.

4. Ensure `grunt browserify` has been run to rebuild the javascript bundle after updating to this Pulsar version.


js/libs/popover.js
```diff
@@ -115,7 +115,21 @@ var $ = require('jquery'),
...
+    $('[data-toggle="popover"], [rel="clickover"]')
+    .on('click', function(e) {
+      e.preventDefault();
+      e.stopPropagation();
+    }).popover();
+  });
+
+  $(document).on('click', function (e) {
+    $('[data-toggle="popover"][data-autoclose="true"], [rel="clickover"]').each(function () {
+        //the 'is' for buttons that trigger popups
```
mikedevelops: can we cache `$(this)` and `$(e.target)` in here

js/libs/popover.js
```diff
@@ -115,7 +115,21 @@ var $ = require('jquery'),
...
+        //the 'has' for icons within a button that triggers a popup
+        if (!$(this).is(e.target) && $(this).has(e.target).length === 0 && $('.popover').has(e.target).length === 0) {                
+            (($(this).popover('hide').data('bs.popover')||{}).inState||{}).click = false  // fix for BS 3.3.6
```
mikedevelops: Should we make this easier to reason about, something like this might work...

```
var $bsPopover = $this.popover('hide').data('bs.popover');

if ($bsPopover !== undefined && $bsPopover.inState) {
    $bsPopover.click = false;
}
```

js/libs/popover.js
```diff
@@ -115,7 +115,21 @@ var $ = require('jquery'),
...
+      e.preventDefault();
+      e.stopPropagation();
+    }).popover();
+  });
+
+  $(document).on('click', function (e) {
+    $('[data-toggle="popover"][data-autoclose="true"], [rel="clickover"]').each(function () {
+        //the 'is' for buttons that trigger popups
+        //the 'has' for icons within a button that triggers a popup
+        if (!$(this).is(e.target) && $(this).has(e.target).length === 0 && $('.popover').has(e.target).length === 0) {
```
mikedevelops: I'm fairly certain this will cover all of these conditions, `closest` checks against the element it's being invoked on and then walks up the tree, so in this instance `e.target` will not be `this` or `.popover` and it will also not be a descendant of those nodes.

```if ($(e.target).closest($(this, '.popover')).length === 0)```

File: js/libs/popover.js [hunk 4]
```diff
@@ -115,7 +115,21 @@ var $ = require('jquery'),
...
+
+  $(document).on('click', function (e) {
+    $('[data-toggle="popover"][data-autoclose="true"], [rel="clickover"]').each(function () {
+        //the 'is' for buttons that trigger popups
+        //the 'has' for icons within a button that triggers a popup
+        if (!$(this).is(e.target) && $(this).has(e.target).length === 0 && $('.popover').has(e.target).length === 0) {                
+            (($(this).popover('hide').data('bs.popover')||{}).inState||{}).click = false  // fix for BS 3.3.6
+        }
+    });
   });
```
mikedevelops: Do we need to add some test coverage?

jamesjacobs: +1

Stanton:
@jamesjacobs @mikedevelops tests added

Stanton:
@jamesjacobs @mikedevelops Awaiting re-review please, we have a major release on the horizon so it would be good to include this at the same time.

PR #892, opened 2018-10-01
New sandboxed styles for XFP conditional logic rules

Stanton:
Adds styles to support an if/and/or/else/elseif structure rather than the workflow style when/and/or/then and draws the connecting lines slightly differently to reduce the number of modifier classes required to try make rule graphs easier to generate with JS.

screenshot 2019-02-07 at 10 51 13

### Example markup

See `rules.html.twig`


stylesheets/_component.rules.scss
```diff
@@ -298,3 +298,306 @@ td.table__form .rules + .form__group:last-of-type {
...
+        }
+    }
+
+    .rule-block + .rule-block,
+    .rule-block--indent > .rule-block {
+        margin-top: -1px;
+
+        // visually distinguish blocks close together
+        .rule-block__header {
+            border-top: 1px solid rgba(255, 255, 255, .5);
```
mikedevelops: super minor but this could be shortened `rgba(white, .5)`

stylesheets/_component.rules.scss
```diff
@@ -298,3 +298,306 @@ td.table__form .rules + .form__group:last-of-type {
...
+        &::after {
+            top: 7px;
+        }
+    }
+
+    .rule-link--and {
+        border-left: 4px solid color(gray, lightest);
+        margin-left: 20px;
+
+        button {
```
mikedevelops: I know we aren't fully BEM compliant at times, but it does feel odd to see a mix of BEM `.thing--modifier` mixed with inheritance and element selectors in the same block.

stylesheets/_component.rules.scss
```diff
@@ -298,3 +298,306 @@ td.table__form .rules + .form__group:last-of-type {
...
+            background-color: lighten(#e8b750, 30);
```
mikedevelops: A few instances of colours in the wild, not sure where we draw the line on converting these to variables, to me it feels as though there should be no colours without variables, so colour changes only ever need to occur in one place framework-wide

stylesheets/_component.rules.scss
```diff
@@ -298,3 +298,306 @@ td.table__form .rules + .form__group:last-of-type {
...
+        margin-left: 20px;
+        padding-left: 20px;
+    }
+
+    .rule-block__header {
+        border: 0;
+        box-shadow: none;
+        color: color(white);
+        font-family: $font-family-regular;
+        font-size: 18.66px; // For AA large text
```
mikedevelops: Perhaps an opportunity to bake the AA compliant sizes into the existing font-size variable set?

stylesheets/_component.rules.scss
```diff
@@ -298,3 +298,306 @@ td.table__form .rules + .form__group:last-of-type {
         width: auto;
     }
 }
+
+/**
+ * The following version 2 was developed to meet the needs of XFP conditional logic rules
+ */
+.rules-container--v2 {
```
jamesjacobs: Small thing but maybe something a little less versioned? `rule-container--logic` perhaps?

stylesheets/_component.rules.scss
```diff
@@ -298,3 +298,306 @@ td.table__form .rules + .form__group:last-of-type {
...
+        border-bottom: 4px solid color(gray, lightest);
+        top: 0;
+    }
+
+    &::after {
+        border-top: 4px solid color(gray, lightest);
+        bottom: 0;
+    }
+
+    .rule-links button {
```
jamesjacobs: I'd BEM the button - `.rule-links__button`.

stylesheets/_component.rules.scss
```diff
@@ -298,3 +298,306 @@ td.table__form .rules + .form__group:last-of-type {
...
+        margin-left: 20px;
+
+        button {
+            background-color: lighten(#e8b750, 30);
+        }
+    }
+
+    .rule-link--else-if {
+        margin-left: -20px;
+        padding-top: 20px;
```
jamesjacobs: alpha order

stylesheets/_component.rules.scss
```diff
@@ -298,3 +298,306 @@ td.table__form .rules + .form__group:last-of-type {
...
+        padding-bottom: 20px;
+
+        &::before {
```
jamesjacobs: alpha order

stylesheets/_component.rules.scss
jamesjacobs: +1 for that

stylesheets/_component.rules.scss
```diff
@@ -298,3 +298,306 @@ td.table__form .rules + .form__group:last-of-type {
...
+
+            @include respond-min($screen-tablet) {
+                color: color(text);
+            }
+        }
+    }
+
+    .rule-block--else-if {
+        &::after {
+            border-top: 4px solid color(gray, lightest);
```
jamesjacobs: alpha order

stylesheets/_component.rules.scss
```diff
@@ -298,3 +298,306 @@ td.table__form .rules + .form__group:last-of-type {
...
+        margin-top: 0;
+
+        .rule-block__header {
+            background-color: #f7d183;
+            color: color(black);
+        }
+    }
+
+    .rule-block--or {
+        background-color: lighten(#9d7ee3, 25);
```
jamesjacobs: lots of colour repeating - could we var these, even if it's in this partial?

stylesheets/_component.rules.scss
```diff
@@ -298,3 +298,306 @@ td.table__form .rules + .form__group:last-of-type {
...
+            @include respond-min($screen-small) {
+                border-top-left-radius: 14px;
+                border-bottom-left-radius: 14px;
+                width: 120px;
+            }
+        }
+    }
+
+    .rule-block__controls {
+        margin-top: .5em;
```
jamesjacobs: We've got a mix of em and px margins in these changes.

views/playground/xfp-rules/rules.html.twig
```diff
@@ -55,424 +55,176 @@
     }}
 {%- endblock -%}
 
+{% set controls = '
+<div class="rule-block__control">
+    <div class="form__group form__group--medium">
+        <label class="control__label hide">Choose trigger</label>
```
jamesjacobs: lots of missing `for`'s here. I know it's a playground so it doesn't matter that much but I'd add them in case someone blindly copies.

stylesheets/_component.rules.scss
jackbentley: Can we get a tool to automate this? Seems trivial?

stylesheets/_component.rules.scss
jackbentley: **em** vs. **px** 🥊 *ding ding* 🔔

stylesheets/_component.rules.scss
Stanton: Because this is an in-flight project I'm going to use `.rule-link .btn` instead, which means I don't have to trigger a round of re-tests for the XFP team.

stylesheets/_component.rules.scss
Stanton: probably something we can add to the linter somehow

stylesheets/_component.rules.scss
Stanton: The ems are overriding/matching form control styles

stylesheets/_component.rules.scss
jamesjacobs: Pretty sure sass linter has this but we've never turned it on due to older code.

views/playground/xfp-rules/rules.html.twig
Stanton: not doing as part of this change

stylesheets/_component.rules.scss
Stanton: already being implemented in-product so will leave as-is for now

stylesheets/_component.rules.scss
```diff
@@ -298,3 +298,338 @@ td.table__form .rules + .form__group:last-of-type {
...
+
+    .rule-block--else-return {
+        background-color: color(gray, lightest);
+        border-color: color(gray, dark);
+        border-radius: 15px;
+
+        .rule-block__header {
+            background-color: color(gray, dark);
+            border-top-left-radius: 14px;
+            border-top-right-radius: 15px;
```
jackbentley: Looks like this should be 14px. Not 15px.

jackbentley:
screen shot 2019-02-13 at 18 02 41
This is still an issue on the very smallest view port width

stylesheets/_component.rules.scss
Stanton: Cheers, fixed!

Stanton:
> screen shot 2019-02-13 at 18 02 41
> This is still an issue on the very smallest view port width

Are you sure it's up to date/cache cleared etc? that should've been fixed by this change:

screenshot 2019-02-13 at 18 41 19

rule

jackbentley:
Resolved the line issue. This was due to having the `rules-container` classes on a `.form__group .controls` element. Adding it as a child `div` solved it! 👍

PR #893, opened 2018-10-01
Remove Wraith tests due to lack of use

### LINKED ISSUE #872: Wraith Issues while prep the Release

Stanton:
We don't use Wraith in any significant capacity, it takes forever to run, baseline results vary from person to person and it's in a broken state at the moment. My current vote is to remove.

Closes #872


nvasileiadis:
Wraith could be fantastically useful if at some point set up correctly in CMS. As there are times we have to go and search CMS-wide for visual changes. Example: the datatables change. There are expected visual changes but sometimes there is just something broken. 

As for Pulsar, because we are always improving and add new components, almost certain is that everytime wraith will complain. Maybe not eliminate it altogether but have it as an extra tool, similar to W3C HTML validator. A separate command and task out of Pulsar workflow.

PR #896, opened 2018-10-05
Make sure quaternary nav works with button links

Stanton:
Fixes a console log issue raised in https://jadultd.atlassian.net/browse/CMS-5669

## Breaking change checklist

| ---- | ---- |

### DIFF
--- a/js/NavMainComponent.js
+++ b/js/NavMainComponent.js
@@ -89,10 +89,10 @@ NavMainComponent.prototype.init = function () {
         var $self = $(this),
             target;
 
-        if ($self[0].hasAttribute('href')) {
+        if ($self.attr('href')) {
             target = $self.attr('href');
         }
-        else if ($self[0].hasAttribute('data-target')) {
+        else if ($self.attr('data-target')) {
             target = $self.attr('data-target');
         }
 
--- a/js/NavMainComponent.js
+++ b/js/NavMainComponent.js
@@ -87,16 +87,23 @@ NavMainComponent.prototype.init = function () {
     // Open quaternary nav on tertiary nav item click
     component.$tertiaryNavLinks.on('click', function (event) {
         var $self = $(this),
-            href = $self.attr('href');
+            target;
 
-        if (href.indexOf('#') !== -1) {
+        if ($self[0].hasAttribute('href')) {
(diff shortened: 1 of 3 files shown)


js/NavMainComponent.js
```diff
@@ -87,16 +87,23 @@ NavMainComponent.prototype.init = function () {
     // Open quaternary nav on tertiary nav item click
     component.$tertiaryNavLinks.on('click', function (event) {
         var $self = $(this),
-            href = $self.attr('href');
+            target;
 
-        if (href.indexOf('#') !== -1) {
+        if ($self[0].hasAttribute('href')) {
```
mikedevelops: I think this `$self[0]` referencing is a bit messy, can we either drop `$(this)` and use `this` or use `$self` exclusively? So in this example instead of accessing the `HTMLElement` with the index, can't we just use `$self.attr('href')` and do away with the `hasAttribute`

Stanton: I mistakenly thought that `if ($self.attr('href')) {` would be problematic, but tests pass and behaviour seems fine

js/NavMainComponent.js
```diff
@@ -87,16 +87,23 @@ NavMainComponent.prototype.init = function () {
...
 
-        if (href.indexOf('#') !== -1) {
+        if ($self.attr('href')) {
+            target = $self.attr('href');
+        }
+        else if ($self.attr('data-target')) {
+            target = $self.attr('data-target');
+        }
+
+        if (target.indexOf('#') !== -1) {
```
mikedevelops: Just a point of discussion here not something that necessarily needs actioning. We still have the ability here to call `.indexOf()` on `undefined` which will throw the same error this ticket is fixing. 

As I see it we have 2 options: 

We could make sure that this code could never throw an error and return inside this function early if `target` is `undefined`. This means that if I implement the nav menu markup incorrectly and omit those attributes, I'll never know it doesn't work until a manually test it. We would effectively be swallowing that implementation error for the sake of it not making it to production.

Or we leave this as is, and put a meaningful error in this block if `target` is `undefined`, we could throw an error a long the lines of `A nav link target requires a href or data-target attribute`.

I think door number 2 would be the friendlier approach for product teams.

PR #899, opened 2018-10-09
Catch cases where a nav link may not have a href or data-target attribute

Stanton:


tests/js/web/NavMainComponentTest.js
```diff
@@ -570,8 +605,10 @@ describe('NavMainComponent', function () {
         });
 
         it('should hide the open tertiary or quaternary nav', function () {
-            expect(this.$navTertiary.hasClass('is-open')).to.be.false;
-            expect(this.$navQuaternary.hasClass('is-open')).to.be.false;
+            setTimeout(() => {
+                expect(this.$navTertiary.hasClass('is-open')).to.be.false;
+                expect(this.$navQuaternary.hasClass('is-open')).to.be.false;
+            }, 100);
```
mikedevelops: Return of the timeouts 😄 , if we're writing asynchronous tests, we need to use Mocha's `done` callback so that the assertion is run as part of the test: (https://mochajs.org/#detects-multiple-calls-to-done)

tests/js/web/NavMainComponentTest.js
```diff
@@ -233,6 +242,32 @@ describe('NavMainComponent', function () {
...
+    describe('clicking on a nav link with a missing data-target', function () {
+        beforeEach(function () {
+            this.navMainComponent.init();
+            this.clickEvent = $.Event('click');
+        });
+
+        it('should throw an error', function () {
+            expect(() => {
+                this.$missingDataTarget.trigger(this.clickEvent);
+            }).to.throw('A nav link must have a href or data-target attribute');
```
mikedevelops: One super minor thing here, these string constants would be good candidates as static "class" properties. So we could perhaps define them on the `NavMainComponent` like so...

```
NavMainComponent.MISSING_ATTR_ERROR = "You broke it";

// throw this string
throw new Error(NavMainComponent.MISSING_ATTR_ERROR);

// assert this string
...to.throw(NavMainComponent.MISSING_ATTR_ERROR);
```

Stanton: Do you mean define them in the actual component, not the test? If so, where?

mikedevelops: yes in the component itself, I normally put them at the top 

```
Foo.GREETING = 'Hello, World!';

function Foo () {}

Foo.prototype.sayGreeting = function () {
	console.log(Foo.GREETING);
}
```

They could go anywhere (outside of a method / block) in that file though.

PR #903, opened 2018-10-16
Allow datatables to accept a custom pageLength

Stanton:
Defaults to 25, allows a custom page length to be set for a specific table, useful when embedding tables in modals where a shorter page length is required.

**Examples**

```
<table class="table datatable" data-page-length="10">...</table>
```

```
{{ 
    html.datatable({
        'data': tableData,
        'data-page-length': '10'
    })
}}
```


Stanton:
The change I pushed yesterday should fix the coverage drop, but CodeCov hasn't updated

Stanton:
% Coverage drop is due to file size increasing

PR #911, opened 2018-11-01
Add sass/browserify compilation to Travis

Stanton:
## Included in this pull request
1. Add grunt tasks to Travis builds
1. Update `grunt-real-favicon` module

## 1. Add grunt tasks to Travis builds
This change includes the `grunt build` task in Travis builds, which runs the following tasks:

```
grunt.registerTask('build', [
    'scsslint',
    'replace',
    'sass:dist_modern',
    'sass:dist_ie',
    'autoprefixer',
    'browserify:dist',
    'copy:dist',
    'compress'
]);
```

Any errors in these tasks will now cause build failures which can be investigated prior to merging.

These changes will increase build times by around 1m 30s, split between running the grunt task and installing the scss-lint gem.

## 2. Update `grunt-real-favicon` module

The `grunt-real-favicon` task started throwing errors, it looks like an API endpoint has moved. The solution to this was to update the version of realFavicon we use.

```
>> Tasks directory "/Users/jamesjacobs/Documents/pulsar/node_modules/grunt-sass-replace/node_modules/grunt-string-replace/tasks" not found.

Running "realFavicon:cms" (realFavicon) task
>> <!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN">
>> <html><head>
>> <title>301 Moved Permanently</title>
>> </head><body>
>> <h1>Moved Permanently</h1>
>> <p>The document has moved <a href="https://realfavicongenerator.net/api/favicon">here</a>.</p>
>> <hr>
>> <address>Apache/2.4.7 (Ubuntu) Server at realfavicongenerator.net Port 80</address>
>> </body></html>
Fatal error: <!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN">
<html><head>
<title>301 Moved Permanently</title>
</head><body>
<h1>Moved Permanently</h1>
<p>The document has moved <a href="https://realfavicongenerator.net/api/favicon">here</a>.</p>
<hr>
<address>Apache/2.4.7 (Ubuntu) Server at realfavicongenerator.net Port 80</address>
</body></html>
```


mikedevelops:
Do we need to run the favicon task as a CI job? It seems like something that is only going to be needed once in a blue moon. 

There's also potentially an argument here that we shouldn't have product favicons within Pulsar at all. This could be a problem for products that need to change a favicon in an emergency but do not have the resources to undertake a (potentially) major pulsar release. Feels a bit tightly coupled.

PR #922, opened 2018-11-12
Fix issue where children of elements triggering popovers do not trigger popovers

jamesjacobs:
Previously clicking a child of an element which has `data-toggle="popover"` did not trigger the popover as expected.

### DIFF
--- a/js/libs/popover.js
+++ b/js/libs/popover.js
@@ -122,9 +122,10 @@ var $ = require('jquery'),
   });
 
   $(document).on('click', function (e) {
+    var $target = $(e.target);
 
     // Ignore popover and clickover triggers and children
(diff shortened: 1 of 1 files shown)


js/libs/popover.js
```diff
@@ -123,8 +123,8 @@ var $ = require('jquery'),
 
   $(document).on('click', function (e) {
 
-    // Ignore popover and clickover triggers
-    if ($(e.target).attr('data-toggle') === 'popover' || $(e.target).attr('rel') === 'clickover') {
+    // Ignore popover and clickover triggers and children
+    if ($(e.target).attr('data-toggle') === 'popover' || $(e.target).parent().attr('data-toggle') === 'popover' || $(e.target).attr('rel') === 'clickover') {
```
mikedevelops: Can we assign `$(e.target)` to a variable here? That way we can avoid creating 3 duplicate jQuery objects, bit of a minor performance optimisation, but they all add up

jamesjacobs: 👍

PR #941, opened 2019-02-04
Close quaternary menu when more icon is clicked a second time

### LINKED ISSUE #930: 'More...' Menu Item Behaves weirdly when clicked
**Describe the bug**
The 'More...' submenu doesn't behave itself. When you click 'More...' and choose a menu item from within it, clicking 'More...' again doesn't replace the opened submenu.

**To Reproduce**
1. Go to https://myaccount.release.xfp.ntn.jadu.net/jadu/xforms2/forms
2. Click on 'More...' in the main nav bar (ensure screen is small enough to show this!)
3. Click an item in 'More...'
4. Click 'More...' again in the sidebar
5. Notice that the 'More...' menu doesn't open, and instead a white line is visible between the main nav and the open menu.

**Expected behavior**
The 'More'...' menu items should be visible.

**Screenshots**
pulsar

 - OS: OSX
 - Browser: Chrome
 - Version 70

**Smartphone (please complete the following information):**
Untested

**Additional context**
I believe this is because the item that is clicked in 'More...' opens over 'More...' instead of replacing it. This can be verified by choosing an item in 'More...' and then closing the new submenu - 'More...' is still expanded.

Stanton:
This change makes sure the quaternary navigation is closed when the more icon is clicked a second time.

### Before
The quaternary navigation stays open.

nav-before

### After
The quaternary navigation is closed.

nav-after

Fixes #930 

## Breaking change checklist

| ---- | ---- |
| require changes to existing unit tests | ✅ No |

### DIFF
--- a/js/NavMainComponent.js
+++ b/js/NavMainComponent.js
@@ -415,7 +415,9 @@ NavMainComponent.prototype.moreIconClickHandler = function ($moreLink) {
     if (component.$navTertiary.find('.nav-list').hasClass('is-active')) {
         $moreLink.attr('aria-expanded', 'false');
         component.$navTertiary.removeClass('is-open');
+        component.$navQuaternary.removeClass('is-open');
         component.$navTertiary.find('.nav-list').removeClass('is-active');
+        component.$navQuaternary.find('.nav-list').removeClass('is-active');
     } else {
         $moreLink.attr('aria-expanded', 'true');
 
--- a/tests/js/web/NavMainComponentTest.js
+++ b/tests/js/web/NavMainComponentTest.js
@@ -462,14 +462,30 @@ describe('NavMainComponent', function () {
             this.$moreIconLink.trigger(this.clickEvent2);
 
             expect(this.$html.find('.nav-tertiary').hasClass('is-open')).to.be.false;
-        });  
+        });
+
+        it('should close the quaternary nav if it is already open', function () {
+            this.$moreIconLink.trigger(this.clickEvent);
+            this.$tertiaryLink.trigger(this.clickEvent);
+            this.$moreIconLink.trigger(this.clickEvent2);
+
+            expect(this.$html.find('.nav-quaternary').hasClass('is-open')).to.be.false;
+        });
 
         it('should remove the is-active class from the tertiary navs active nav-list if its already open', function () {
             this.$moreIconLink.trigger(this.clickEvent);
             this.$moreIconLink.trigger(this.clickEvent2);
 
             expect(this.$html.find('.nav-tertiary .nav-list').hasClass('is-active')).to.be.false;
-        });  
+        });
+
+        it('should remove the is-active class from the quaternary navs active nav-list if its already open', function () {
+            this.$moreIconLink.trigger(this.clickEvent);
+            this.$tertiaryLink.trigger(this.clickEvent);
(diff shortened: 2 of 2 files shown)


js/NavMainComponent.js
```diff
@@ -415,7 +415,9 @@ NavMainComponent.prototype.moreIconClickHandler = function ($moreLink) {
     if (component.$navTertiary.find('.nav-list').hasClass('is-active')) {
         $moreLink.attr('aria-expanded', 'false');
         component.$navTertiary.removeClass('is-open');
+        component.$navQuaternary.removeClass('is-open');
         component.$navTertiary.find('.nav-list').removeClass('is-active');
+        component.$navQuaternary.find('.nav-list').removeClass('is-active');
     } else {
```
mikedevelops: ⚠️ Micro Optimisation Alert ⚠️ – We could cache a lot of these `.find(...)` queries ahead of time. We could store a reference to these nodes on load which would save performing them each time the event fires. However this comes with the caveat that the menu markup would not change during it's lifespan (which I think probably _is_ the case)

I totally understand if we do not want to change it to implement this ☝️

Stanton: `53 occurrences`

tests/js/web/NavMainComponentTest.js
```diff
@@ -462,14 +462,30 @@ describe('NavMainComponent', function () {
...
             expect(this.$html.find('.nav-tertiary').hasClass('is-open')).to.be.false;
-        });  
+        });
+
+        it('should close the quaternary nav if it is already open', function () {
+            this.$moreIconLink.trigger(this.clickEvent);
+            this.$tertiaryLink.trigger(this.clickEvent);
+            this.$moreIconLink.trigger(this.clickEvent2);
+
+            expect(this.$html.find('.nav-quaternary').hasClass('is-open')).to.be.false;
```
mikedevelops: Should we chuck in an ```expect(this.$html.find('.nav-quaternary').hasClass('is-open')).to.be.true;``` after the first click events? Wouldn't this test pass if the functionality that adds `'is-open` is broken? Admittedly that would (hopefully) fail elsewhere. Might be nice to catch that here also though, so it shouts very loudly if something is broken.

tests/js/web/NavMainComponentTest.js
```diff
@@ -462,14 +462,30 @@ describe('NavMainComponent', function () {
...
             expect(this.$html.find('.nav-tertiary .nav-list').hasClass('is-active')).to.be.false;
-        });  
+        });
+
+        it('should remove the is-active class from the quaternary navs active nav-list if its already open', function () {
+            this.$moreIconLink.trigger(this.clickEvent);
+            this.$tertiaryLink.trigger(this.clickEvent);
+            this.$moreIconLink.trigger(this.clickEvent2);
+
+            expect(this.$html.find('.nav-quaternary .nav-list').hasClass('is-active')).to.be.false;
```
mikedevelops: ☝️

js/NavMainComponent.js
mikedevelops: 53, yikes. Maybe a candidate for some love if we ever get into any FE performance issues.

Stanton:
@mikedevelops Added requested changes.

PR #943, opened 2019-02-07
Fix inputs not meeting WCAG 2.1 colour contrast

jamesjacobs:
This branch changes contains the following fixes to meet GUI elements WCAG 2.1 contrast requirements:

- All inputs now have a darker border
- Toggle switch off background is slightly darker
- Range slider has a slightly lighter track background colour but a darker border colour. Focus styles have also being included and IE9 fallbacks added (previously missing).
- Select2's now have the same bg colour as other inputs.

Before:
screenshot 2019-02-07 at 16 49 11

After:
screenshot 2019-02-07 at 16 48 28


stylesheets/_component.range-slider.scss
```diff
@@ -13,12 +13,12 @@
...
     // Slider thumb
     &::-webkit-slider-thumb {
-        margin-top: -5px;
+        margin-top: -7px;
         appearance: none;
         background: color(primary);
         border: 0;
         border-radius: 50%;
-        cursor: pointer;
+        cursor: ew-resize;
```
Stanton: Is this appropriate? I'd expect `grab` perhaps?

jamesjacobs: I think I'll revert this. Was trying to highlight that the thumb could be moved left and right, though that cursor is obviously more relevant to resizing things.

jamesjacobs: I've changed the cursor to grab @Stanton

PR #945, opened 2019-02-11
Initial attempt to standardise datatable options into an extendable method

### LINKED ISSUE #944: Add ability to override datatable init() method
**Is your feature request related to a problem? Please describe.**
XFP need to deploy a datatable which uses AJAX as the data source. Currently there's no way for a datatable to bypass the Pulsar init() method while retaining datatable styling.

**Describe the solution you'd like**
Either a simple `data-init="false"` option which would stop a given table from being initialised, and/or a new method which would allow developers to fetch the desired DT options which Pulsar uses for datatable functionality.

**Additional context**
screenshot 2019-02-11 at 09 05 34

Stanton:
This change creates a new `getDatatableOptions()` method which returns an object containing Pulsar's standard options. This can be extended to override or add new options.

## DataTable options

Where possible you should use Pulsar's default options to initialise your datatable, if you're using the `html.datatable()` twig helper, or a regular HTML table with the `.datatable` class, these options should be intialised for you automatically.

If you need to have more control over your table and want to opt-out of Pulsar's automatic initialisation method, you must supply the `data-init="false"` attribute on the `table` element.

```twig
{{ 
    html.datatable({
        'id': 'myDatatable',
        'data-init': 'false',
        'data': data
    }) 
}}

or

<table id="myDatatable' class="datatable table--horizontal" data-init="false">...</table>
```

You can access the Pulsar datatable options via the `getDatatableOptions()` method, you should pass a reference to the table in the DOM so that the method can access any data attributes that may have been set.

```javascript
// Store a reference to your desired table
let $myDatatable = $('#myDatatable');

// Fetch the default Pulsar options
const pulsarDatatableOptions = pulsar.pulsarUI.getDatatableOptions($myDatatable);

// This presumes you're requiring the relevant DataTable files in your code
let table = this.DataTable(pulsarDatatableOptions);
```

### Extending the default DataTable options

You can replace/extend values of the default options to suit your DataTable needs.

```javascript
// Store a reference to your desired table
let $myDatatable = $('#myDatatable');

// Fetch the default Pulsar options
const pulsarDatatableOptions = pulsar.pulsarUI.getDatatableOptions($myDatatable);

// Example to show us changing the value of `stateSave`
const customDatatableOptions = $.extend({}, pulsarDatatableOptions, {
    stateSave: true
});

// This presumes you're requiring the relevant DataTable files in your code
let table = this.DataTable(customDatatableOptions);
```

Closes #944


js/PulsarUIComponent.js
```diff
@@ -35,10 +35,77 @@ PulsarUIComponent.prototype.init = function () {
...
+        },
+        select: select,
+        selector: 'td.table-selection',
+        stateSave: false,
+        style: 'multi'
+    };
+
+    return options;
+};
+
```
mikedevelops: This method looks as though it could be static. This could potentially be useful in environments where a developer does not want to instantiate the PulsarUIComponent but _does_ want the Pulsar datatable options.

It should be as simple as changing  ```PulsarUIComponent.prototype.getDatatableOptions = function ($table) {``` to ```PulsarUIComponent.getDatatableOptions = function ($table) {```

We'd then need to change any references to this method to ```PulsarUIComponent.getDatatableOptions(...)```

js/PulsarUIComponent.js
```diff
@@ -52,62 +119,15 @@ PulsarUIComponent.prototype.initTables = function () {
...
-        if ($this.data('page-length')) {
-            pageLength = $this.data('page-length');
-        }
+        var $this = $(this);
 
-        if ($this.data('select') === false) {
-            dom = '<"dataTables_top"irf><"dataTables_actions"T><"dt-disable-selection"t><"dataTables_bottom"p>';
-            select = false;
-        }
+        const datatableOptions = component.getDatatableOptions($this);
```
mikedevelops: A quick example if that options method were to become static, this line would become 
```
const datatableOptions = PulsarUIComponent.getDatatableOptions($this);
```

js/PulsarUIComponent.js
```diff
@@ -52,62 +119,15 @@ PulsarUIComponent.prototype.initTables = function () {
...
-            },
-            responsive: {
-                details: {
-                    type: 'column'
-                }
-            },
-            select: select,
-            stateSave: false
-        });
+        let table = $this.DataTable(datatableOptions);
```
mikedevelops: This could be a `const`

js/PulsarUIComponent.js
```diff
@@ -127,52 +147,27 @@ PulsarUIComponent.prototype.initDataTables = function () {
...
-                "infoEmpty": 'No items',
-                "infoFiltered": " (filtered from _MAX_ items)",
-                "zeroRecords": "No items matched your filter, please clear it and try again",
-                "search": "Quick filter"
-            },
             select: select,
-            stateSave: false
         });
+        
+        let table = $this.DataTable(horizontalOptions);
```
mikedevelops: This could be a `const`

js/PulsarUIComponent.js
Stanton: Done.

js/PulsarUIComponent.js
Stanton: Done.

js/PulsarUIComponent.js
Stanton: Done.

js/PulsarUIComponent.js
Stanton: Done.

Stanton:
@mikedevelops Suggestions implemented, thank you.

PR #948, opened 2019-02-21
Table detail accessibility and focus management improvements

jamesjacobs:
Various accessibility improvements, including:

1. Added `role="dialog"` and  `aria-modal="true"` to table detail panel
2. Trapping focus inside of the open panel
3. Returning focus to triggering element
4. Attaching panel to `main` landmark if one is present
5. ESC to close

No breaking changes.


js/TableDetailComponent.js
```diff
@@ -18,20 +18,33 @@ class TableDetailComponent {
...
-            '       <button type="button" class="close table-detail__header-close" data-table-detail-close-panel aria-hidden="true">&times;</button>' +
+            '       <button type="button" class="close table-detail__header-close" data-table-detail-close-panel aria-label="Close"><span aria-hidden="true">&times;</span></button>' +
             '       <h1 class="table-detail__title" data-table-detail-panel-title>Detail</h1>' +
             '   </div>' +
             '   <div class="table-detail__body" data-table-detail-panel-body></div>' +
             '</div>'
-        );
+            ),
+            $elementToAppendTo,
+            $containerHasMainElement = this.$html.find('main'),
```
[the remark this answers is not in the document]
Stanton: Spotted you do this in the `trapFocus()` method, so would be a little consistency

js/TableDetailComponent.js
```diff
@@ -99,6 +134,66 @@ class TableDetailComponent {
...
+        // Remove previously bound listener
+        this.$html.off('keydown', this.boundKeydownListener);
+    }
+
+    /**
+     * Listen for keyboard navigation, trap tabbing within the panels focusable elements
+     * @param {jQuery} $focusableElements - Collection of focusable elements in the details panel
+     */
+    keydownListener ($focusableElements, event) {
+        let keyCode = event.keyCode || event.which;
```
mikedevelops: The general consensus is that `let` should be used in contexts where the variable will be reassigned. In situations where a variable is declared and assigned once, it should probably be a `const`.

There is an interesting argument / quite sweary rant (https://jamie.build/const) that you should only ever use `let` because javascript "constants" can be mutated, I can understand the principle, but I'm not sure I agree with it. I think the distinction between `let` and `const` is useful because when skimming code it's easy to reason about which variables may end up reassigned.

js/TableDetailComponent.js
```diff
@@ -18,22 +18,39 @@ class TableDetailComponent {
...
 
         this.$table = this.$html.find('[data-table-detail-table]');
 
-        // Add backdrop and detail panel if UI contains a table detail pattern
-        if (this.$table.length) {
-            this.$html.find('body')
-                .append('<div class="table-detail-backdrop"></div>')
-                .append($panelHtml);
+        // Check panel hasn't already been added, in case of a datatable re-initting this component on paginated page draw
+        if ($panelAlreadyExists.length === 0) {
```
Stanton: shorter with `if (!$panelAlreadyExists.length) {` ?

js/TableDetailComponent.js
```diff
@@ -88,6 +118,15 @@ class TableDetailComponent {
 
         // Open panel
         this.$detailPanel.addClass('table-detail--open');
+
+        // Make elements focusable again
+        this.$detailPanel
+            .find('a[href], area[href], input, select, textarea, button, iframe, object, embed, [tabindex], *[contenteditable]')
```
Stanton: Duplicate lists of element types here, would be betters as a var/const

js/TableDetailComponent.js
jamesjacobs: Personally I prefer the explicitness of `=== 0`. I can change if we really feel this is an issue.

js/TableDetailComponent.js
jamesjacobs: Done

Issue #951, opened 2019-03-05
Main content can be mis-aligned by mobile swipe


jamesjacobs:
**Describe the bug**
On mobile devices, swiping left reveals a large white area.

**To Reproduce**
1. Load a Pulsar UI on mobile
2. Swipe left
3. Main content moves off screen.

This is due to change in #946 

**Expected behavior**
Content should not move off screen.

**Screenshots**
![image](https://user-images.githubusercontent.com/756393/53817195-d7b4c400-3f5c-11e9-8a87-9460f9b03eee.png)

**Smartphone (please complete the following information):**
 - Device:iPhone simulator
 - OS: iOS
 - Browser: Safari

PR #952, opened 2019-03-05
Fix mobile swipe issue

jamesjacobs:
Fixes issue reported in #951
UI can no longer be swiped to the left.
Original fix for search bar being obscured after jumping to a page fragment still works.


jamesjacobs:
Browser tested fix and original fragment fix in:
- Chrome
- Firefox
- Safari
- IE11-9
- iOS safari

Issue #787, opened 2018-03-22
Tab order when triggering secondary navigation needs improvement


jamesjacobs:
**Current behavior:**
![secondary nav tab order](https://user-images.githubusercontent.com/756393/37783319-1908b920-2ded-11e8-949a-fbd753a2f21f.gif)

When you open the secondary navigation via keyboard, you have to tab through the rest of the primary nav items before being able to access the opened secondary nav links.

**Desired behavior:**
The next tab would take you to the secondary nav items. 

I've explored a few options, all of which would require breaking changes. The best options appear to be repeating the secondary navigation within each primary navigation `li` so natural tab order is preserved. This will require JS and CSS changes.

Stanton:
Closed by #1257

PR #955, opened 2019-03-19
WIP - nav focus improvements

jamesjacobs:
**WIP - not complete and untested**

Attempting to fix the issues mentioned in #787 by changing the way the secondary navigation works. Currently a separate div exists `nav-secondary` which is a sibling of `nav-primary`. This MR removes that and adds a separate `nav-secondary` after each button that triggers secondary navigation. This allows us to rely on source order for focus order. 

A user can tab through the primary nav as normal. If a secondary nav is opened then due to the DOM location the next tab will focus the first interactive element in the opened secondary navigation.

This branch also contains focus management improvements to the open/close interactions of more button, `nav-tertiary` and `nav-quaternary`.

I've not yet changed the nav helper whilst there are still issues to be looked at so the MR contains a hardcoded navigation which will need removing once the changes have made to the twig helper.

Remaining todo/issues:

1. mobile secondary navigation is broken due to changes.
2. when `nav-quaternary` is open, you can `shift + tab` and cycle through `nav-tertiary` links. As these are blocked by `nav-quaternary` being open, suggest we remove them from tab order whilst `nav-quaternary` is open or close `nav-quaternary` when one receives focus.
3. slight spacing/overlap styling issue with the more button and the open `nav-tertiary`.
4. update twig nav helper.
5. Fix/add new JS tests.
6. browser test.

We'll also need to remove tab-indexes on elements in the toolbar and search to make these changes effective. Otherwise focus order will be broken.


Stanton:
@jamesjacobs Still a WIP or abandoned?

Stanton:
@jamesjacobs Do you have any sense of what remains to be done on this PR to get it out of WIP?

jamesjacobs:
@Stanton Picking this up again today

jamesjacobs:
Closing due to nav focus management fixes introduced in release 11/12

PR #966, opened 2019-04-29
Make pre/appended content be constrained by grid modifiers

Stanton:
This change makes prepended/appended content hidden on overflow so that grid widths affect the total with of the input including any prepended/appended content. 

## Before
Screenshot 2019-04-29 at 09 53 13

## After
Screenshot 2019-04-29 at 09 54 53

Change affects:
* form.date
* form.password
* form.range
* form.select
* form.select2
* form.text
* form.time

### DIFF
--- a/stylesheets/_component.forms.scss
+++ b/stylesheets/_component.forms.scss
@@ -397,6 +397,7 @@ fieldset {
 .form__group.form__group .input-group .select2.select2,
 .form__group.form__group .input-group .form__control.form__control {
     flex: 1 1 auto;
+    overflow: hidden;
     width: 1%; // needed to stop overflow, when option has a long string on < desktop
 
     @include ie-lte(9) {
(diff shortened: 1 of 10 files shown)


stylesheets/_component.forms.scss
```diff
@@ -397,6 +397,7 @@ fieldset {
 .form__group.form__group .input-group .select2.select2,
 .form__group.form__group .input-group .form__control.form__control {
     flex: 1 1 auto;
+    overflow: hidden;
```
jamesjacobs: Are you sure this is needed? In my quick tests the issue was fixed with `min-width: 0` and `overflow: hidden` on `.input-group-addon`

Stanton: select2s completely broke at small widths without this, I'll provide screenshot

Stanton: Before
Screenshot 2019-04-30 at 10 15 14

After
Screenshot 2019-04-30 at 10 13 20

jamesjacobs: 👍

PR #968, opened 2019-04-30
Color contrast of dropzone helptext now meet AA contrast

### LINKED ISSUE #967: Dropzone help text doesn't meet AA colour contrast
**Describe the bug**
Dropzone help text doesn't meet AA colour contrast. Current contrast ratio is 4.41 out of a required 4.5.

jamesjacobs:
Color contrast of dropzone helptext now meet AA contrast when used with form__control class.

Fixes #967


stylesheets/_component.dropzone.scss
```diff
@@ -93,7 +93,7 @@ $dropzone-icon-rotation: 30deg;
 }
 
 .dropzone__help {
-    color: color(text, help);
```
Stanton: > Search: `color(text, help)`
> 6 results in 4 files 

arguably, the colour value for this token should updated instead, making any/all help text instances AA compliant.

jamesjacobs: General help text is AA compliant. This is when the dropzone is used with `form__control` and therefore has the `form__control` off white bg colour.

Stanton: aha ok, I retract my change request!

PR #973, opened 2019-05-01
Fixed issue where esc listener would fire when table detail not present

### LINKED ISSUE #972: Table detail adds ESC key listener, which triggers when table detail isn't present
**Describe the bug**
TableDetailComponent.js attaches an ESC keypress event listener that incorrectly triggers when the pattern isn't being used and when the panel isn't open resulting in a console error.

Needs a check adding to ensure the panel is present and open.

**To Reproduce**
1. Visit any CXM UI using the latest Pulsar release (not yet released in CXM)
2. Press the ESC key

**Expected behavior**
The event listener should only trigger when the panel is open.

jamesjacobs:
Fixes #972


js/TableDetailComponent.js
```diff
@@ -88,7 +92,7 @@ class TableDetailComponent {
 
         // Close ESC button
         this.$html.on('keydown', (event) => {
-            if (event.keyCode === 27) {
+            if (event.keyCode === 27 && this.$tableDetailBackdrop.hasClass('in')) {
```
mikedevelops: Really minor here I'm not suggesting we change it unless you feel really compelled to do so. In instances like this it might be nicer to determine if something is active by accessing some internal state in our Component, like an isActive getter or something. The problem with hanging off the presentation class "in" is aside from mixing our concerns we can not change the implementation of "in" now without having to factor this service also. 

That might feel like a pain if we assumed we were making a purely presentational (HTML & CSS) change like changing "in" to "is-in" (as a contrived example), but now we also need to change javascript components and their tests.

jamesjacobs: @mikedevelops Updated to do this. Totally agree about separating concerns.

Issue #971, opened 2019-05-01
Choice helper (v2) doesn't add required attributes to inputs


Stanton:
**Describe the bug**
The `form.choice()` v2 helper doesn't add the `required` and/or `aria-required` attributes to the radio inputs when `required: true` is provided as an option.

**To Reproduce**
```
{{
        form.choice({
            'class': '',
            'label': 'Choice radios',
            'required': true,
            'options': [
                {
                    'label': 'One',
                    'value': 'one',
                    'name': 'one'
                },
                {
                    'label': 'Two',
                    'value': 'two',
                    'name': 'one'
                },
                {
                    'label': 'Three',
                    'value': 'three',
                    'name': 'one'
                }
            ],
            'help': 'Help text to give more information about expected input'
        })
    }}
```

It's possible to add the `required: true` to each individual item in the `options` array, which will result in the addition of the required attributes, however it will also duplicate the required asterisk on the label for each radio options, and is a non-ideal method of supplying the attributes anyway.

**Expected behavior**
The inclusion of `required: true` as a top-level option on the form.choice() helper should result in each input defined by `options` having the `required` boolean attribute, and the `aria-required="true"` attribute applied to the `<input>` element.

PR #975, opened 2019-05-02
Improve how required choice helpers are marked up

Stanton:
Previously, using the `required` option on form.choice helpers wouldn't mark the fields as required. 

This change adds the `required aria-required="true"` attributes to the choice inputs (checkbox and radio) and makes sure the required asterisk label is only added to the top-level label.

## Before
Screenshot 2019-05-02 at 14 05 36

## After
Screenshot 2019-05-02 at 14 08 22

```
<div class="form__group form-choice">
    <label class="control__label">foo&nbsp;<span class="required-indicator" data-toggle="tooltips" title="required">*</span></label>
    <div class="controls">
        <label for="form_field_0" class="control__label">
            <input 
                id="form_field_0" 
                name="form[field]" 
                value="0" 
                checked="checked" 
                required aria-required="true"
                type="radio" 
                class="form__control radio" 
            /><span class="form-choice__label">Foo</span></label>
        <label for="form_field_1" class="control__label">
            <input 
                id="form_field_1" 
                name="form[field]" 
                value="1" 
                required aria-required="true"
                type="radio" 
                class="form__control radio" 
            /><span class="form-choice__label">Bar</span></label>
        <label for="form_field_2" class="control__label">
            <input 
                id="form_field_2" 
                name="form[field]" 
                value="2" 
                required aria-required="true"
                type="radio" 
                class="form__control radio" 
            /><span class="form-choice__label">Baz</span></label>
    </div>
</div>
```


Stanton:
Closes #971

jamesjacobs:
Looking at the example given above, now would be a good time to group choice options with a fieldset/legend to avoid WCAG `1.3.1 Info and Relationship` and `3.3.2 Labels or Instructions` `form elements are not grouped` failures.

Stanton:
> Looking at the example given above, now would be a good time to group choice options with a fieldset/legend to avoid WCAG `1.3.1 Info and Relationship` and `3.3.2 Labels or Instructions` `form elements are not grouped` failures.

This is a larger scope change not related to #971 which will need to be considered separately.

PR #977, opened 2019-05-16
Update to jQuery 3.x

### LINKED ISSUE #666: Move away from Bower

Stanton:
Major refactoring of javascript tooling to streamline package management and update the core jQuery version.

## Main changes
* Update jQuery dependency to 3.x
* Update any packages for jQuery 3.x support
* Move all Bower dependencies to NPM
* Remove all references to Bower
* Update javascript components to reference NPM dependencies
* Remove all oldIE / libs / theme references from base template
* Update `.gitattributes`, `.gitignore` and `makefile`

### Styles
* Susy grid
* Projector palette and themes (and grunt tasks relating to theming)

### Javascript
* History.js
* Livereload-js
* NWMatcher
* PhantomCSS
* Selectivizr
* Svgeezy
* Tinycon

## Breaking change checklist

| ---- | ---- |
| require changes to `main.js` | ⚠️ YES |
| require changes to `index.js` | ⚠️ YES |
| require changes to `pulsar.scss` | ⚠️ YES |
| require changes to `gruntfile.js` | ⚠️ YES |
| require changes to `package.json` (not including dev dependencies) | ⚠️ YES |
| require changes to `base.html.twig` (or other core views like header/footer) | ⚠️ YES |
| require new arguments to be passed to a js component | ⚠️ YES |
| require changes to existing unit tests | ⚠️ YES |

Closes #666


mikedevelops:
@Stanton `InputCloneService` is all patched up now

bower.json
```diff
@@ -13,7 +13,7 @@
     "chartist": "^0.9.8",
     "font-awesome": "4.7.0",
     "html5shiv": "latest",
-    "jquery": "1.x",
+    "jquery": "3.x",
```
mikedevelops: Do we need this if we are committing jQuery?

mikedevelops:
Could we potentially reduce the repo bloat here by just committing the dist version of jquery (if we absolutely have to commit it)

bower.json
Stanton: First steps, keeping the change footprint small

Stanton:
> Could we potentially reduce the repo bloat here by just committing the dist version of jquery (if we absolutely have to commit it)

That would make the way we currently use Bower a bit non-standard, the `.gitattributes` removes the 'bloat' from the product, and there's still the plan on removing our dependency on Bower or bundling jQuery at all, just not yet as part of this ticket.

mikedevelops:
In my opinion installing a dependency with bower, committing all the files, then selectively ignoring some (which has broken in the past) is the more fragile non-standard way to handle them. 

If we were to commit only the dist/compiled/bundled libraries, we'd remove the uncertainty around which directories and files to ignore through .gitattributes. We'd always know we have everything we need without having to maintain a long list of things we think we don't need.

However, that that said...

With the work required and the time it will take on products to test the changes of the jQuery3 upgrade, I honestly think it would be less work to simply remove the committed dependency from Pulsar (add it to our npm deps for local development and the lexicon) and then have products install their own versions of jQuery at a product level and start to remove this idea that Pulsar manages your UI _and_ your dependencies. I think that would be less work from a maintainability standpoint in both the products and in Pulsar.

Stanton:
> In my opinion installing a dependency with bower, committing all the files, then selectively ignoring some (which has broken in the past) is the more fragile non-standard way to handle them.
> 
> If we were to commit only the dist/compiled/bundled libraries, we'd remove the uncertainty around which directories and files to ignore through .gitattributes. We'd always know we have everything we need without having to maintain a long list of things we think we don't need.
> 
> However, that that said...
> 
> With the work required and the time it will take on products to test the changes of the jQuery3 upgrade, I honestly think it would be less work to simply remove the committed dependency from Pulsar (add it to our npm deps for local development and the lexicon) and then have products install their own versions of jQuery at a product level and start to remove this idea that Pulsar manages your UI _and_ your dependencies. I think that would be less work from a maintainability standpoint in both the products and in Pulsar.

We're on the same side with this, as I said in my last comment, this **will** happen, just hasn't happened yet on this 'still a work in progress' ticket ;)

tests/harness/common.js
```diff
@@ -11,6 +11,8 @@ require('babel-polyfill');
 // Expose jQuery globals
 window.$ = window.jQuery = $;
 
+console.log($.fn.jquery);
+
```
jackbentley: Debug code

views/pulsar/layouts/base.html.twig
```diff
@@ -188,5 +188,6 @@
 
 {% block javascript %}{% endblock %}
 {% block tab_javascript %}{% endblock %}
+<script src="https://code.jquery.com/jquery-migrate-3.1.0.js"></script>
```
jackbentley: Should this still be here?

jackbentley:
Looks like you got around to removing jQuery from bower? Could we do the same for jQuery-ui and vide as they've both been updated in this PR?

Stanton:
That will be a separate PR outside of the change to jQuery, but likely in the same release. Can you ticket it please?

Stanton:
Actually nah we'll just do it in this one

js/PulsarFormComponent.js
```diff
@@ -1,10 +1,11 @@
 'use strict';
 
 var $ = require('jquery'),
-    TimePickerComponent = require('./TimePickerComponent');
+TimePickerComponent = require('./TimePickerComponent');
```
jamesjacobs: Not important but we normally tab these var declarations out like prior to your change.

js/area/signin/signin.js
```diff
@@ -303,7 +304,7 @@ SignInComponent.prototype.switchPanel = function (panelClass) {
 	newPanel
 		.attr('aria-hidden', 'false')
 		.find('[tabindex]')
-		.removeAttr('disabled')
+		.prop('disabled', false)
```
jamesjacobs: This would give us a mix of `.prop('disabled', false)` and `attribute('disabled', 'disabled')` on `oldPanel` is that intentional? See L316.

stylesheets/_deprecated.styles.scss
```diff
@@ -468,42 +468,42 @@ $rich-list-item-text-selected-color:  color(white) !default;
...
-.grid-col {
-    margin-top: gutter();
-
-    &:last-child {
-        @include respond-min($screen-phone) {
-            @include last;
-        }
-    }
-}
+// $susy: (
```
jamesjacobs: Any reason for commenting these out rather than removing?

tests/js/web/FilterBarComponentTest.js
```diff
@@ -61,8 +61,7 @@ describe('FilterBar component', function () {
 	});
 
 	afterEach(function () {
-        delete $.fn.select2;
-        delete $.fn.popover;
+		this.$body.empty();
```
jamesjacobs: funky tabbing

tests/js/web/FilterBarComponentTest.js
```diff
@@ -201,14 +200,14 @@ describe('FilterBar component', function () {
 		});
 
 		it('should prevent the default behavior', function () {
-			this.$popoverFilterLink = this.$container.find('[data-filter-id="colour"]');
+			this.$popoverFilterLink = this.$container.find('.filter-bar__list [data-filter-id="colour"]');
```
jamesjacobs: more funky tabbing and below 👇

views/pulsar/layouts/base.html.twig
```diff
@@ -44,27 +36,8 @@
...
     <meta name="theme-color" content="#ffffff">
 
-    <!--[if IE 7]><link rel="stylesheet" type="text/css" href="{{ cssPath ~ 'pulsar-ie7' ~ themeFile ~ '.min.css' }}" /><![endif]-->
-    <!--[if IE 8]><link rel="stylesheet" type="text/css" href="{{ cssPath ~ 'pulsar-ie8' ~ themeFile ~ '.min.css' }}" /><![endif]-->
-    <!--[if IE 9]><link rel="stylesheet" type="text/css" href="{{ cssPath ~ 'pulsar-ie9' ~ themeFile ~ '.min.css' }}" /><![endif]-->
-    <!--[if gt IE 9]><!-->
-    <link rel="stylesheet" type="text/css" href="{{ cssPath ~ 'pulsar' ~ themeFile ~ '.css' }}" />
+    <link rel="stylesheet" type="text/css" href="{{ cssPath ~ 'pulsar.css' }}" />
     <link rel="stylesheet" type="text/css" href="{{ cssPath ~ 'lexicon.css' }}" />
-    <!--<![endif]-->
```
jamesjacobs: 🔥

tests/js/web/FilterBarComponentTest.js
Stanton: fine in code

tests/js/web/FilterBarComponentTest.js
Stanton: fine in code

views/pulsar/layouts/base.html.twig
Stanton: ?

stylesheets/_deprecated.styles.scss
Stanton: removed

js/DisableUiComponent.js
```diff
@@ -57,16 +57,16 @@ DisableUiComponent.prototype.enable = function (target) {
 
         // Enable form elements
         $this.find(FORM_ELEMENTS)
-            .unbind('click', preventDefaultAndStopPropagation)
+            .off('click', preventDefaultAndStopPropagation)
             .removeClass('disabled')
-            .removeAttr('disabled');
+            .prop('disabled', false);
```
jamesjacobs: We're using `prop` here in the `enable` method but setting it as an attribute in the `disable` method - see L32 (`.attr('disabled', 'disabled');`). Using a mix may cause issues. We should either set/unset the attribute or property not a mix of both.

js/area/signin/signin.js
```diff
@@ -312,7 +313,7 @@ SignInComponent.prototype.switchPanel = function (panelClass) {
 	oldPanel
 		.attr('aria-hidden', 'true')
 		.find('[tabindex]')
-		.attr('disabled', 'disabled')
+		.prop('disabled', 'disabled')
```
jamesjacobs: This should be `.prop('disabled', 'true')`.

stylesheets/pulsar.scss
```diff
@@ -16,19 +16,19 @@ $ie-version: 12 !default;
 
 // Third party plugins
 // scss-lint:disable all
-@import '../libs/normalize-css/normalize';
-@import '../libs/susy/sass/susy';
+@import '../node_modules/normalize.css/normalize';
+//@import '../libs/susy/sass/susy';
```
jamesjacobs: Remove?

PR #979, opened 2019-05-17
Update to Twig 2.x

### LINKED ISSUE #978: Update Twig dependency to 2.x
Current Twig dependency is `1.21`.

To facilitate Jadu platform updating to Symfony 3.4 the supported Twig version will be updated to 2.x (2.10.0 at the time of writing).

Stanton:
This change updates the Twig dependency from 1.x to 2.x, and resolves all issues reported by unit tests and manual testing of the Pulsar Lexicon.

Closes #978 

## Summary of breaking changes encountered

* Updated Travis to use PHP7
* Added explicit `@import` rules to include helpers like `html`, `form`, `nav`, `flash` etc...
* Changed `replace` filter as now requires an array or traversable instead of a string, eg `|replace({'a': 'b'})` instead of `|replace('a', 'b')`

## Breaking change checklist

| ---- | ---- |
| require changes to `base.html.twig` (or other core views like header/footer) | 🔴 YES |
| require markup changes to maintain functionality | 🔴 YES (Numerous changes to Twig views) |
| require changes to existing unit tests | 🔴 YES |


Stanton:
@aallott Just noticed you have a branch open which updates the version in composer, wasn't aware you were picking this up...

views/pulsar/components/tab.html.twig
```diff
@@ -7,26 +7,39 @@
...
-    {% endblock %}
+{% if block('actions_left') is defined %}
+    {% set _actions_left = block('actions_left') %}
 {% endif %}
 
-{% set _actions_left = block('actions_left') %}
-{% set _actions_right = block('actions_right') %}
+{% if block('actions_right') is defined %}
+    {% set _actions_right = block('actions_right') %}
+{% endif %}
```
jackbentley: We don't need to put these into variables. We can just check if the block is defined below and then output the `block('actions_right')` directly.

views/pulsar/components/tab.html.twig
```diff
@@ -7,26 +7,39 @@
...
+    {% set _sidebar = block('tab_sidebar') %}
+    {% if _sidebar is not empty %}
+        {% set _hasSidebar = true %}
+    {% endif %}
 {% endif %}
 
-{% set _sidebar = block('tab_sidebar') %}
-{% if _sidebar is not empty %}
-    {% set _hasSidebar = true %}
+{% if block('tab_js') is defined %}
```
jackbentley: Might be worth just removing support for `tab_js` block since we're making a breaking change release anyway?

Stanton:
Heard very recently that Twig 2 is no longer on the roadmap for Jadu products. You'll need to speak to your friendly neighbourhood product owner if you need it.

Stanton:
Reopening as part of stack update work

Stanton:
I'm not doing any merges or updates to this branch until Stellar can confirm which base release they need/want this to apply to, whether that's Pulsar 11 or 12.

Issue #983, opened 2019-05-21
Accessibility insights report on pagination colour contrast in XFP


markbjadu:
Pagination Elements must have sufficient colour contrast

1. Go to 'Xforms Pro > and select or create a new form'
2. Click on 'Advanced tab', create and save some rules. 
3. Go back to advanced tab and run 'Accessibility insights'
4. See the report

"Element has insufficient color contrast of 1.85 (foreground color: #bebebe, background color: #ffffff, font size: 10.5pt, font weight: normal). Expected contrast ratio of 4.5:1"

 - OS: [mac OS 10.14.5]
 - Browser [chrome]
 - Version [74.0.3729.157]

![Screenshot 2019-05-21 at 10 11 45](https://user-images.githubusercontent.com/6906173/58083629-447d4900-7bb1-11e9-882d-b949b951f772.png)

markbjadu:
Shatabdi mentioned that the forwards and previous only need to be visible if there are extra pages. Can they be hidden if there is just one page the << >>? @Stanton.

Stanton:
Basic tables will need their pagination routines updating in-product to match the update recommendations. Keeping ticket open for the time being.

Stanton:
No longer an issue after Purple

PR #987, opened 2019-06-03
Datatables pagination a11y

jamesjacobs:
This branch addresses a number of issues with the datatables pagination listed below:

1. Disabled links (such as "previous" on when on page 1) were shown to the user, could be focused and failed AA contrast. These have now been hidden both visually and from screen readers when not relevant. For example, there is no previous link when on page 1 and no next link when on the final page.

2. Link text was non descriptive to screen reader users. `aria-label`'s have now been added tp pagination links, for example `aria-label="page 1"` along with `aria-current="true"` where a link refers to the currently displayed page.

3. Pagination was shown when no records or only one page of records were present. Pagination is now only disabled when there are enough records to cause pagination.

4. For large amounts of records, First and Last links weren't present. These have been added with appropriate `aria-label`'s.

5. The mark up datatables generates isn't the most semantic. Pagination links are now wrapped in a `nav` element with an accessible name.

6. The record length picker was incorrectly missing from all datatables. This has been re-added.

7. The record length picker was missing the default Pulsar orange focus indicator and relying on the browser default style. This has been fixed.

8. The tab order was incorrect, after tabbing from the last focusable element in a table, focus would move to the record length picker and not the pagination. This has been fixed.

I've also updated the lexicon non datatable pagination examples with more accessible markup and fixed an overflow issue.

**Before:** 

image

**After:**

image

image

image

**Browser tested in:**

- [x] Chrome
- [x] Firefox
- [x] Safari
- [x] Edge
- [x] IE11-9

**Screen reader tested in:**

- [x] JAWS 2018 IE11
- [x] JAWS 18 IE11
- [x] NVDA IE11
- [x] Mac VO Safari and Chrome

Addresses #983

### DIFF
--- a/js/PulsarUIComponent.js
+++ b/js/PulsarUIComponent.js
@@ -42,6 +42,7 @@ PulsarUIComponent.getDatatableOptions = function ($table) {
     let dom = '<"dataTables_top"Birf><"dataTables_actions"T>t<"dataTables_bottom"pl>',
         langEmptyTable = 'There are currently no items to display',
         pageLength = 25,
+        lengthChange = false,
         select = {
             className: 'dt-row-selected',
             style: 'multi',
@@ -54,6 +55,10 @@ PulsarUIComponent.getDatatableOptions = function ($table) {
 
     if ($table.length && $table.data('page-length')) {
         pageLength = $table.data('page-length');
+    }
+
+    if ($table.length && $table.data('length-change')) {
+        lengthChange = $table.data('length-change');
     }
 
     if ($table.length && $table.data('select') === false) {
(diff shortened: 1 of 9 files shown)


js/PulsarUIComponent.js
```diff
@@ -64,31 +64,66 @@ PulsarUIComponent.getDatatableOptions = function ($table) {
...
+                paginate: {
+                    first:    'First page',
+                    previous: 'Previous page',
+                    next:     'Next page',
+                    last:     'Last page'
+                }
+            }
         },
-        lengthChange: false,
+        lengthChange: true,
```
Stanton: This will need flagging up in the release notes as a change to expected behaviour, we may need to add the ability for it to be disabled via a data-attribute.

js/PulsarUIComponent.js
```diff
@@ -3,6 +3,9 @@
 var $ = require('jquery'),
     StickyScrollBarComponent = require('./StickyScrollBarComponent');
 
+const { initComplete } = require('./DataTables/datatablesInitComplete');
```
Stanton: Oh this is cheeky, I like it.

PR #991, opened 2019-06-05
Remove leftover files related to Wraith

Stanton:
These are no longer required and should've been removed as part of #893


makefile
```diff
@@ -36,14 +36,6 @@ build:
 	@ sudo gem install scss_lint -v ${SASSLINTVER} --no-ri --no-rdoc --no-user-install
 	@ echo "\n${CHECK} Done"
 
-	@ echo "${HR}\nInstalling PhantomJS...${HR}\n"
```
jamesjacobs: We use PhantomJS for JS tests- or is this something different?

jamesjacobs: @mikedevelops ?

jamesjacobs: Nope we use chrome headless for tests, ignore me.

PR #1000, opened 2019-07-04
Add feature to pre-fill a repeatable component 

mikedevelops:
Adds the ability to pre-fill the Repeater Component with values on load.

The public API hasn't changed so should just be a patch version I think. This is required for an upcoming XFP story where they are using the component for the first time.

If anyone has time to take it down the accessibility audit that would be greatly appreciated, I'm fairly certain this went out before we started auditing...


Stanton:
Merged into master and released as 10.3.1

PR #1004, opened 2019-07-18
Add iso and locale formatting to date helper

jackbentley:
POC of enabling the date helper to post back its data value in the HTML5 ISO format.

Additionally, there is a new format called `locale` which will automatically determine the format to display based on the browser locale.

### DIFF
--- a/js/PulsarFormComponent.js
+++ b/js/PulsarFormComponent.js
@@ -90,7 +90,7 @@ PulsarFormComponent.prototype.initDatePickers = function () {
 
     datepickers.each((index, element) => {
         let $element = $(element);
-        let dateFormat;
+        let dateFormat = 'DD/MM/YYYY';
         let formatKey = element.getAttribute('data-format');
         let locale = element.getAttribute('data-locale');
 
@@ -108,8 +108,6 @@ PulsarFormComponent.prototype.initDatePickers = function () {
                 case 'reverse':
                     dateFormat = 'YYYY/MM/DD';
                     break;
-                default:
-                    dateFormat = 'DD/MM/YYYY';
             }
         }
 
--- a/js/PulsarFormComponent.js
+++ b/js/PulsarFormComponent.js
@@ -89,6 +89,7 @@ PulsarFormComponent.prototype.initDatePickers = function () {
     const datepickers = this.$html.find('[data-datepicker="true"]');
 
     datepickers.each((index, element) => {
+        let $element = $(element);
         let dateFormat;
         let formatKey = element.getAttribute('data-format');
         let locale = element.getAttribute('data-locale');
@@ -113,19 +114,19 @@ PulsarFormComponent.prototype.initDatePickers = function () {
         }
 
         // Initialize pikaday with the correct date format
-        $(element).pikaday({ format: dateFormat });
+        $element.pikaday({ format: dateFormat });
 
         // Initialize placeholder attribute based on the date format
-        $(element).attr('placeholder', dateFormat.toLowerCase());
+        $element.attr('placeholder', dateFormat.toLowerCase());
 
         if (element.hasAttribute('data-iso') && element.hasAttribute('name')) {
-            let elementName = $(element).attr('name');
+            let elementName = $element.attr('name');
             let hiddenElement = $('<input>', {'type': 'hidden', 'name': elementName});
 
-            $(element).removeAttr('name');
-            $(element).after(hiddenElement);
+            $element.removeAttr('name');
+            $element.after(hiddenElement);
 
-            $(element).change(function() {
+            $element.change(function() {
                 let parsed = moment($(this).val(), dateFormat);
                 let ISOTime = parsed.toISOString(true);
 
--- a/js/PulsarFormComponent.js
+++ b/js/PulsarFormComponent.js
@@ -1,8 +1,11 @@
 'use strict';
 
 var $ = require('jquery'),
-    TimePickerComponent = require('./TimePickerComponent');
+    TimePickerComponent = require('./TimePickerComponent'),
+    moment = require('moment');
 
+require('moment/locale/en-au');
+require('moment/locale/en-gb');
 require('../libs/pikaday/plugins/pikaday.jquery');
 require('../libs/select2/dist/js/select2.min');
 require('../libs/spectrum/spectrum');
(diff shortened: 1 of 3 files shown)


js/PulsarFormComponent.js
```diff
@@ -86,31 +89,56 @@ PulsarFormComponent.prototype.initDatePickers = function () {
     const datepickers = this.$html.find('[data-datepicker="true"]');
     let defaultDateFormat = 'DD/MM/YYYY';
 
+    moment.locale(window.navigator.userLanguage || window.navigator.language);
```
Stanton: I'm not sure we'll want to automatically change the locale based on user language, we'll quite likely need this to be an explicit user preference, or system preference, in a higher level UI.

js/PulsarFormComponent.js
```diff
@@ -86,31 +89,56 @@ PulsarFormComponent.prototype.initDatePickers = function () {
...
             case 'reverse':
-                defaultDateFormat = 'YYYY/MM/DD';
+                dateFormat = 'YYYY/MM/DD';
+                break;
+            case 'locale':
+                dateFormat = moment.localeData().longDateFormat('L');
                 break;
             default:
-                defaultDateFormat = 'DD/MM/YYYY';
+                dateFormat = 'DD/MM/YYYY';
```
Stanton: `dateFormat = defaultDateFormat` ?

js/PulsarFormComponent.js
```diff
@@ -86,31 +89,56 @@ PulsarFormComponent.prototype.initDatePickers = function () {
...
 
         // Initialize placeholder attribute based on the date format
-        $(element).attr('placeholder', defaultDateFormat.toLowerCase());
+        $(element).attr('placeholder', dateFormat.toLowerCase());
+
+        if (element.hasAttribute('data-iso') && element.hasAttribute('name')) {
+            let elementName = $(element).attr('name');
+            let hiddenElement = $('<input>', {'type': 'hidden', 'name': elementName});
+
+            $(element).removeAttr('name');
```
Stanton: cache `$(element)` ?

js/PulsarFormComponent.js
jackbentley: This is based on the browser's language setting. The same one which decides the `Accept-Language` request header. I'd assume this setting is set based on the system's language setting when the browser is installed.

js/PulsarFormComponent.js
jackbentley: You need to create a local variable. Otherwise, you're altering the default. I think there might be a bug at the moment given multiple inputs:

```twig
{{ 
    form.date({
        'format': 'US',
    })
}}

{# This input would be in the US format, I think #}
{{ 
    form.date()
}}
```

jackbentley: Perhaps not, because of the default. However, the local variable is needed for the event handler so it has a solid reference that won't change

js/PulsarFormComponent.js
Stanton: Sorry by 'system' I meant our software. The software should, by default, provide a consistent date format for all users which may then have the ability to be changed as a user preference, I just wanted to check that we're not inheriting the default locale based on the users location or OS preferences, it needs to be initially defined by the software constants.

Public XFP forms are a slight outlier in this situation where it's more appropriate for dates to be in the user's expected locale, but this needs to not be the default for the rest of the software, where more consistency is required.

jackbentley: Interesting. I'd argue that the app should adapt to the user's setting where possible. That's based on my own opinion and the whole idea of "progressive" web apps. Would be interesting to get some research on the topic?

Stanton: This is based on the context of our software (the traditional use of 'people working at an organisation' rather than members of the public, which as I mentioned, is a non-default use case here). Dates need to be consistent across the software for members of the same organisation, for a single basic example: consider a call centre agent talking to a third-party service provider, talking about the date where an action should be taking place, it would be easy for confusion to happen if they're quoting dates to each other but unaware their preferences differ, and DD/MM vs MM/DD is being compared.

There's the ability/need for timezone/locale information to be used when displaying dates also, though that's outside the scope of this PR.

js/PulsarFormComponent.js
mikedevelops: I'm not sure if I'm following, feels like what @Stanton is suggesting makes the most sense here? Keeping the `dateFormat` variable but then assigning it the value of the `defaultDateFormat` (line 90) in this default case

js/PulsarFormComponent.js
jackbentley: Thinking outside the box, perhaps the actual solution to that problem would be to not display dates in that way. For example, always use the written month instead of the number - `12 June` over `12/06`. That way it's obvious no matter what. Say, for example, I live in the UK, and I move to the US to work - I would 100% get confused even when trying my best to stick to the correct date. If we always display in the word then this is avoided.

Stanton: Sure, however date formats are also a customer preference that we'd need to work with.

jackbentley: Back on topic, however: given the reason I opened this PR is due to the differences server-side and client-side of rendering and evaluation of date-related things; it would help if this setting was able to accept an actual locale (such as `en_au`). That would mean I don't have to do any weird processing of the locale in the application in order to use it in Pulsar.

js/PulsarFormComponent.js
jackbentley: Apologies! 🤦‍♂ I got confused. I agree @Stanton is correct now I've re-read this!

js/PulsarFormComponent.js
jackbentley: I've reworked this to accept a `locale` option instead 👍

js/PulsarFormComponent.js
```diff
@@ -84,33 +87,56 @@ PulsarFormComponent.prototype.initTimePickers = function () {
...
+                    dateFormat = 'YYYY/MM/DD';
+                    break;
+                default:
+                    dateFormat = 'DD/MM/YYYY';
+            }
         }
 
         // Initialize pikaday with the correct date format
-        $(element).pikaday({ format: defaultDateFormat });
+        $element.pikaday({ format: dateFormat });
```
Stanton: Can you add a check in here, while the helper will now always set a `data-locale` attribute there may be hardcoded HTML date pickers in production which won't have either `data-format` or `data-locale` which may result in `dateFormat` being undefined, if I'm not mistaken?

jackbentley: 👍 Have made the default the UK format which can be changed via the attributes

js/PulsarFormComponent.js
```diff
@@ -84,33 +87,54 @@ PulsarFormComponent.prototype.initTimePickers = function () {
...
+            $element.removeAttr('name');
+            $element.after(hiddenElement);
+
+            $element.change(function() {
+                let parsed = moment($(this).val(), dateFormat);
+                let ISOTime = parsed.toISOString(true);
+
+                if (ISOTime !== null) {
+                    ISOTime = ISOTime.substr(0, 10);
+                }
```
mikedevelops: Could we use Moment's format method here instead of the substr? Taking it a step further we could have that format an option? So we'd be able to configure the value sent to the server when using the component perhaps.

jackbentley: I use the `.toISOString()` method from moment before the string manipulation. The whole point is that it posts back in the ISO format - in the same way an HTML5 input does.

I can change the string manipulation for one of these though if that's more comfortable?
https://stackoverflow.com/a/39736368/1600814

jackbentley: Although there are two other questions around this:
1) Does Pulsar support a `type="datetime"` with pikaday? If so, this needs to be aware of that.
2) In the `.toISOString()` I set the `keepOffset` flag to  `true` - I don't know if this is correct or not.

mikedevelops: On the first point, that SO answer is what I'm suggesting, particularly this `date.utc().format('HH:mm:ss');` as we can have the HTML5 format as the default, but then make it extensible by having an option to set that format string (HH:mm:ss), so when we innevitably need a non HTML5 standard date we can conifigure that per component.

On omitting the `true` argument to `toISOString` it's a tricky one... as we're stripping all the useful information out of the timestring (thanks HTML5!) we'd have to couple the understaning of the timezone between the client and server in the component itself. I'd suggest making this configurable as part of the options also, so we can decide on an implementation basis.

The pickaday question I'm not sure on, one to check for sure ⏲

Stanton: > 1. Does Pulsar support a `type="datetime"` with pikaday? If so, this needs to be aware of that.

No, I don't believe we do. There may well be datetime pickers in the wild, but not provided by Pulsar, apart from the new Symfony forms branch, but that's not Pikaday.

jackbentley: In that case, it seems reasonable to me to use `date.format('YYYY-MM-DD');`. Unsure how/if the `.utc()` is strictly necessary as we're only dealing with dates I don't think we care about the timezone as we need to literally say which day not which hour.

I can't see any reasoning to make the format configurable. We should be standardising on the HTML5 format so that we are able to easily replace the text input with the date input for accessibility when the time comes.

mikedevelops: Good point on the `utc()` I blindly copied that for the SO you linked, I was getting mixed up there with dates and datetimes. UTC can almost certainly be omitted.

In regards to the configurable format, I would still have it as a configurable option, with maybe some accompanying best practice in the documentation. I agree that it _should_ follow the HTML5 spec.

js/PulsarFormComponent.js
```diff
@@ -84,33 +87,54 @@ PulsarFormComponent.prototype.initTimePickers = function () {
...
-        $(element).pikaday({ format: defaultDateFormat });
+        $element.pikaday({ format: dateFormat });
 
         // Initialize placeholder attribute based on the date format
-        $(element).attr('placeholder', defaultDateFormat.toLowerCase());
+        $element.attr('placeholder', dateFormat.toLowerCase());
+
+        if (element.hasAttribute('data-iso') && element.hasAttribute('name')) {
+            let elementName = $element.attr('name');
+            let hiddenElement = $('<input>', {'type': 'hidden', 'name': elementName});
```
jackbentley: The hidden element should be added in the Twig. The current implementation has issues when passing a value into the HTML as a default.

Stanton:
@jackbentley Is this draft PR still a work in progress or abandoned?

Stanton:
Closing due to staleness and lack of test cases to validate whether this POC is valid.

PR #1006, opened 2019-07-24
Stop tooltips in CMS hiding the tooltip parent on leave

Stanton:
So... I'm not entirely sure why CMS behaves differently given the same markup, but it hides the tooltip parent on leave rather than the actual tooltip.

Making the leave method explicitly hide the tooltip element fixes in CMS, and behaviour in Pulsar is not changed.

## Before
tooltip-before

## After
tooltip-after

### DIFF
--- a/js/libs/tooltip.js
+++ b/js/libs/tooltip.js
@@ -119,11 +119,13 @@ var $ = require('jquery');
     var self = obj instanceof this.constructor ?
       obj : $(obj.currentTarget)[this.type](this.getDelegateOptions()).data('bs.' + this.type)
 
+    var $tip = this.tip()
+
     clearTimeout(self.timeout)
 
     self.hoverState = 'out'
-
-    if (!self.options.delay || !self.options.delay.hide) return self.hide()
+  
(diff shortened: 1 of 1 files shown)


jackbentley:
This is happening because CMS _still_ use [prototype.js](http://prototypejs.org/) which adds some weird event handlers when `hide()` is called, causing the element to hide.

In XFP, we have a file called "hide interceptor" which disables this functionality of prototype. I'd have assumed CMS would have something similar. There will be many other similar issues like this without a general fix. 

The absolute best (and highly recommended solution) is to stop using prototype.js - it only powers all the old legacy pages.

Stanton:
> The absolute best (and highly recommended solution) is to stop using prototype.js - it only powers all the old legacy pages.

I don't have the ability to remove prototype at this point, we could possibly add the hide interceptor if you could let me know where to find it, though do you think this is needed given this small PR resolves my issue?

jackbentley:
I'd be concerned that there are various other things which also give the same effect and it's not actually a bug with Pulsar.

PR #1025, opened 2019-09-26
Fix issue preventing DataTables from adopting horizontal scroll behaviour

Stanton:
1. Stop using legacy `bAutoWidth` option since we updated our DT version some time ago, this resolves some issues with width calculation
2. The default behaviour will still be to collapse columns, unless the `.table--horizontal` class is applied
3. When a table is being initialised with the `horizontalOptions`, this will now switch off `responsive` (which is responsible for collapsing columns) and switch on the `scrollX` option to allow scrolling

**Scrolling**
dt-before

**Collapse**
dt-colalpse


stylesheets/_component.datatables.scss
```diff
@@ -83,6 +83,8 @@ a.dt-button {
 
     td,
     th {
+        overflow: hidden;
```
Stanton: DT does some funky stuff when scrolling is enabled, the `thead` and `tbody` are moved into tables entirely separate from each other and the column widths are calculated by DT javascript. The existing `thead` is hidden from from view, but the sortable arrows were still visible. This hides them.

Stanton:
Screenreader issue: The `scrollX` recreates the `thead` and hides the original, however our checkbox is still currently interactive.

dt-sr

stylesheets/_component.datatables.scss
```diff
@@ -404,3 +404,10 @@ th.sorting_disabled:hover {
         }
     }
 }
+
+// Prevent interactive elements in the original head from taking focus on tab
+.dataTables_scrollBody thead tr[role="row"] {
+    visibility: collapse;
```
Stanton: Stop checkboxes in the original hidden `thead` being interactive.

PR #1031, opened 2019-10-08
Fix issue causing modules with no pages to be incorrectly unselected

Stanton:
CMS team reported an issue with the module page component, the following behaviour was not being correctly triggered:

```
Given I have a module with no sub-pages
When I select the main module checkbox
Then the CRUD controls should all be checked
```

**Before**
ezgif-2-13e2b449a55c

**After**
ezgif-2-2113828c1441-1

This change also restores indeterminate styling which was not originally added when we restyled the checkboxes.


Stanton:
Changed base to be merged into `cms-a11y-develop` branch

Issue #1035, opened 2019-10-30
Disabled link functionality references an invalid attribute


Stanton:
https://github.com/jadu/pulsar/blob/develop/js/PulsarUIComponent.js#L26

**Describe the bug**

The code which prevents disabled links from being clickable expects `<a href="#link" disabled>` which is an invalid attribute.

**Expected behavior**
A discussion is required to make sure we're providing a method of disabling links which doesn't contravene WCAG 2.1.

**Additional context**
Twig helpers will use a common `disabled: true` helper option, this should continue to work with whatever outcome we decide upon.

PR #1037, opened 2019-11-01
Improve disabled links javascript

Stanton:
Given the following code is present on page load:

```
<!-- HTML -->
<a href="#link" class="is-disabled">my link</a>

<!-- TWIG -->
{{
    html.link({
        'label': 'my link',
        'disabled': true
    })
}}
```

The `initDisabledLinks` method, invoked on page load, will transform this into:

```
<a class="is-disabled" aria-disabled="true" tabindex="-1" role="button" data-href="#link">my link</a>
```

Disabled links are currently styled like this:
Screenshot 2019-11-01 at 14 08 22


jamesjacobs:
@Stanton I'd have thought we'd have done this in the `link` twig macro/helper, was there a reason you decided on a purely JS solution?

One small thing, we need to remove focus styles from the link which still appear if you click:

Screenshot 2019-11-06 at 15 20 57

Tested in VO/NDA and JAWS 18/2018 - works as expected.

Stanton:
It's rare that the link helper is used over a regular <a> elemt and we need to trigger preventDefault anyway so this solution makes sure disabled links (which NEED is to be disabled) are also using the right combination of attributes we specify.

Stanton:
Closes #1035

PR #1076, opened 2019-11-19
Fix the twig deprecation issue while CMS upgrades to Symfony 3.4

Stanton:
Fixes Twig warnings and errors which will be triggered due to changes in Symfony 3.4.


views/pulsar/components/tab.html.twig
```diff
@@ -8,26 +8,39 @@
...
+{% endif %}
+
+{% set _actions_left = null %}
+{% set _actions_right = null %}
+{% if block('actions_left') is defined %}
+    {% set _actions_left = block('actions_left') %}
+{% endif %}
+{% if block('actions_right') is defined %}
+    {% set _actions_right = block('actions_right') %}
+{% endif %}
```
jackbentley: Could've written these as:
```twig
{% set _actions_left = block('actions_left') is defined ?  block('actions_left') : null %}
```

PR #1081, opened 2019-11-20
Make panel body overflow when contents exceed max-height

### LINKED ISSUE #1061: Check overflow behaviour on panel body
**Describe the bug**
There's a scratch rule in CSS which sets `overflow-y` to prevent scrollbars appearing when workflow review comments are large enough to exceed the 20em max-height.

```
.panel__body {
    overflow-y: auto;
}
```

Check this rule is required, if so, implement in Pulsar and remove the scratch rule from CSS.

Stanton:
This is required as a fallback for CMS where workflow comments may exceed the max-height of a panel. It's rare that this should be triggered.

panel

CMS scratch rule can be removed when product updates to a Pulsar version which includes this PR.

Closes #1061


stylesheets/_component.panels.scss
```diff
@@ -19,6 +19,7 @@
 
     &__body {
         max-height: 20em;
```
jamesjacobs: I know this is existing, but why do we need to specify a max-height here? Without it we could remove the overflow issue completely.

stylesheets/_component.panels.scss
```diff
@@ -19,6 +19,7 @@
...
+        overflow-y: auto;
```
jamesjacobs: There's some additional work required to make scrolling containers keyboard accessible. See: https://developer.paciellogroup.com/blog/2016/02/short-note-on-improving-usability-of-scrollable-regions/

Note: this is subjective depending on if the scrolling container has focusable elements within it.

stylesheets/_component.panels.scss
Stanton: Currently it's not desirable for the panel to exceed a set height when the body content is variable in length.

Stanton:
Updated the panel helper to make panel actions a sibling of body rather than a child, both markup methods are supported in styles though.

This gets around the focusable elements in actions being outside of the visible area and makes them always available.

panel-scroll

stylesheets/_component.panels.scss
jamesjacobs: I think we should add an additional option to `html.panel` for this behaviour, `scroll` maybe? That could be used to add a overflow modifier class for the scroll and also to add `tabindex="0" to the panel body, otherwise there is no way for a keyboard only user to scroll the content of the overflowing panel body (unless they are using Firefox, which I think is the only browser that currently makes block level elements with an overflow focusable via the keyboard).

views/lexicon/tabs/panels.html.twig
```diff
@@ -171,4 +171,50 @@
         })
     }}
 
+    <div class="panel with-timeline u-margin-top-none">
```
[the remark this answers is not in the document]
Stanton: Copied over from CMS

Issue #1078, opened 2019-11-19
Select elements have invalid append/prepend attributes.


jamesjacobs:
Using `form.select()` or `form.select2()` with a prepended or appended button option results in invalid attributes on the select element:

```
{{
        form.select({
            'id': 'select-30',
            'label': 'Prepended button',
            'prepend': html.button({ 'label': 'foo' }),
            'prepend_type': 'button',
            'options': [
                {
                    'label': 'Value one',
                    'value': 'value1'
                },
                {
                    'label': 'Value two',
                    'value': 'value2'
                }
            ]
        })
    }}
```

Results:
```
<select id="select-30" prepend_type="button" class="form__control select"><option value="value1">Value one</option><option value="value2">Value two</option></select>
```

Need to exclude `prepend_type` and `append_type` within the `select` and `select2`helpers.

Issue #1074, opened 2019-11-19
form.content() with the disable option results in invalid disabled attribute


jamesjacobs:
When the disabled option is used with `form.content` it results in the `<p>` having an invalid `disabled` attribute.

```
<p disabled  class="is-disabled">My value</p>
```

Issue #1073, opened 2019-11-19
form.date() adds invalid format attribute to input


jamesjacobs:
form.date() adds an invalid "format" attribute to the input.
```
<input format="US" id="date-US" placeholder="mm/dd/yyyy" data-datepicker="true" data-format="US" type="text"  class="form__control" />
```
Need to be excluded within the helper/macro.

jackbentley:
#1004 potentially related? Maybe not

Issue #1072, opened 2019-11-19
form.compound adds invalid type attribute to certain inputs


jamesjacobs:
Currently, form.compound adds an invalid attribute to certain elements. For example:

```
<select id="unit-0" type="select"  class="form__control u-width-auto select"><option value="sec">Seconds</option><option value="min">Minutes</option></select>
```
The type attributes needs removing for certain non 'input' types (select, textarea and other none `input` types).

PR #1089, opened 2019-11-25
Form helper option/attribute fixes

### LINKED ISSUE #1079: Form helpers with common options can sometimes produce invalid attributes
For example, setting a `required` or `placeholder` option on `form.range()` will result in the input having an invalid `required` or `placeholder` attribute.

Issues found in helpers:
- `form.range`: should ignore/exclude the `required` and `placeholder` options from the input.
- `form.file`: should ignore/exclude the `placeholder` options from the input.

jamesjacobs:
This branch fixes various issues where certain form helper options resulted in invalid attributes being applied to inputs.

This includes:

`form.range`: excludes the `required` and `placeholder` attributes
`form.file`: excludes the `placeholder` attribute
`form.select` / `form.select2`: excludes the `prepend_type` and `append_type` attribute
`form.content`: excludes the `disabled` attribute
`form.date`: excludes the `format` attributes
`form.compound`: excludes the `type` attribute when input type is select, select2 or textarea

Resolves #1079 #1078 #1074 #1073 #1072


views/pulsar/v2/helpers/form.html.twig
```diff
@@ -859,7 +862,7 @@ Option      | Type   | Description
 autofocus   | bool   | Whether the field should have input focus on page load
 disabled    | bool   | Stops the element from being interactive if true
 form        | string | Specific one or more forms this label belongs to
-format      | string | Options for this can be `default`, `US` or `reverse`
+data-format | string | Options for this can be `default`, `US` or `reverse`
```
Stanton: Checked XFP and the `format` attribute is in use, may need to switch back to `format` and convert that to `data-format` inside the helper.

jamesjacobs: Reverted the comment change. The `format` option is still used to set the format, this is passed to `data-format` within the help. The key change here is that `format` is excluded from the markup (which was an invalid attribute).

PR #1097, opened 2019-11-29
Implement sticky sidebar as a Pulsar component

### LINKED ISSUE #1058: Move sticky sidebar support over to Pulsar
**Describe the bug**

XFP FormBuilder includes a sticky sidebar which greatly improves the experience when editing longer forms. This feature should be brought over to Pulsar and documented.

**Scratch rule**
```
@media screen and (min-width: 992px) {
    .tab__inner.is-sticky .tab__content {
        margin-left: calc(370px + 1%);
        max-width: calc(100% - 370px);
    }

    .tab__inner.is-sticky .tab__settings {
        max-width: 370px;
        max-height: 100%;
        overflow-y: auto;
        position: fixed;
        top: 0;
    }
}
```

**JS**
```
'use strict';

(function($) {
    function sticky() {
        
        var windowTop = $(window).scrollTop(),
            $container = $('.tab__container.has-settings .tab__inner'),
            $settings = $('.tab__container.has-settings .tab__settings'),
            containerTop = $container.offset().top;

        if ($('.tab__container.has-settings .tab__content').outerHeight() > $('.tab__container.has-settings .tab__settings').outerHeight()) {
            if (windowTop > containerTop) {
                $container.addClass('is-sticky');
            } else {
                $container.removeClass('is-sticky');
            }
        } else {
            $container.removeClass('is-sticky');
        }
    }

    sticky();

    $(window).on('scroll resize', sticky);

})(jQuery);
```
**Behaviour**

sticky

Stanton:
Expected behaviour:

```
Given there is a settings panel layout
And the component is initialised
And the content area is taller than the sidebar panel
When the page is scrolled so the sidebar panel would be past the top of the viewport
Then the sidebar panel should be `stuck` to the top of the viewport
```

* Both areas should be independently scrollable to ensure content is accessible
* May need to add tabindex to scrollable areas

sticky-scroll

Closes #1058


Stanton:
ezgif-6-ef9b332a5719

Touch scrolling not working properly on iPad, most of the scrolling here (apart from the initial scroll) was being attempted from the sidebar.

js/StickySidebarComponent.js
```diff
@@ -0,0 +1,60 @@
...
+     * if no longer required
+     */
+    sticky () {
+        const $container = this.$html.find('.tab__container.has-settings .tab__inner');
+
+        if (!$container.length) {
+            return false;
+        }
+
+        if ((this.$html.find('.tab__container.has-settings .tab__content').outerHeight() > this.$html.find('.tab__container.has-settings .tab__settings').outerHeight()) && ($(window).scrollTop() > $container.offset().top)) {
```
jamesjacobs: Rather than thew global `window` here `$(window)` could we use `$(this.window)`.

tests/js/web/StickySidebarComponentTest.js
```diff
@@ -0,0 +1,85 @@
...
+    let componentWithoutHTML;
+    let componentWithoutWindow;
+
+	beforeEach(() => {
+        $window = $('<div></div>');
+        window = $window[0];
+        $body = $('body');
+        
+		$layoutHTML = $(
+            '<div class="tab__container has-settings">' +
```
jamesjacobs: Funky tabbing

tests/js/web/StickySidebarComponentTest.js
```diff
@@ -0,0 +1,85 @@
...
+			'   	</section>' +
+			'	    <main class="tab__content">' +
+			'		</main>' +
+			'	</div>' +
+			'</div>'
+		).appendTo($body);
+	});
+
+	afterEach(() => {
+        $body.empty();
```
jamesjacobs: funky tabbing

tests/js/web/StickySidebarComponentTest.js
```diff
@@ -0,0 +1,85 @@
...
+    });
+
+    describe('init()', () => {
```
jamesjacobs: Lots of funky tabbing

tests/js/web/StickySidebarComponentTest.js
Stanton: fine in code

js/StickySidebarComponent.js
```diff
@@ -0,0 +1,60 @@
...
+     * 
+     * If content is taller than settings, and the viewport has been scrolled 
+     * enough then add the required sticky class to the container, and remove it
+     * if no longer required
+     */
+    sticky () {
+        const $container = this.$html.find('.tab__container.has-settings .tab__inner');
+
+        if (!$container.length) {
+            return false;
```
mikedevelops: why return false here?

Stanton: Stop any errors being thrown if sticky() is used but the correct layout is not present

js/StickySidebarComponent.js
```diff
@@ -0,0 +1,60 @@
...
+     * initial sticky calculation and binds to resize & scroll events
+     */
+    init () {
+        if (typeof this.$html === 'undefined' || !this.$html) {
+            throw new Error('$html must be passed to StickySidebarComponent');
+        }
+
+        if (typeof this.window === 'undefined' || !this.window) {
+            throw new Error('window must be passed to StickySidebarComponent');
+        }
```
mikedevelops: we know that `this.$html` & `this.window` are declared, so these undefined checks should just be 
```
this.$html === undefined
this.window === undefined
```

also I think the `|| !this.$html` check is confusing, what are we asserting here exactly?

Stanton: I'm asserting that I copied it from StickyScrollbarComponent 😅

js/StickySidebarComponent.js
```diff
@@ -0,0 +1,60 @@
...
+     * if no longer required
+     */
+    sticky () {
+        const $container = this.$html.find('.tab__container.has-settings .tab__inner');
+
+        if (!$container.length) {
+            return false;
+        }
+
+        if ((this.$html.find('.tab__container.has-settings .tab__content').outerHeight() > this.$html.find('.tab__container.has-settings .tab__settings').outerHeight()) && ($(this.window).scrollTop() > $container.offset().top)) {
```
mikedevelops: I found this quite hard to read, could we maybe break it down and assign these expressions to variables so that the `if` statement is easier to reason about, I was thinking something like

```
const isContentBiggerThanSettings = this.$html.find(...).outerHeight() > ect ect ect
const isScrolledPastContainer = ect ect ect

if (isContentBiggerThanSettings && isScrolledPastContainer) { ... }
```

probably not the best names for things, but might make it easier if there's a bug

js/StickySidebarComponent.js
mikedevelops: I understand the early return, but why return the value `false` ? should this not just be 

```
if (!...) {
    return;
}
```

js/StickySidebarComponent.js
mikedevelops: haha! I suggest we remove it, as it is effectively asserting that if you pass in an argument that is falsy then throw an error (which is kind of fine), but what if I pass in an argument that is wrong but is truthy?

js/StickySidebarComponent.js
```diff
@@ -0,0 +1,63 @@
...
+     * 
+     * If content is taller than settings, and the viewport has been scrolled 
+     * enough then add the required sticky class to the container, and remove it
+     * if no longer required
+     */
+    sticky () {
+        const $container = this.$html.find('.tab__container.has-settings .tab__inner');
+
+        if (!$container.length) {
+            return false;
```
jamesjacobs: No need to return `false` just `return` to exit.

stylesheets/_layout.settings.scss
```diff
@@ -468,3 +468,128 @@
...
+    }
+}
+
+/**
+ * Sticky sidebar
+ */
+
+@media screen and (min-width: 992px) {
+    .tab__inner.is-sticky .tab__content {
+        left: calc(370px + 1%);
```
jamesjacobs: What's the 1% here for?

stylesheets/_layout.settings.scss
```diff
@@ -468,3 +468,128 @@
...
+        position: relative;
+    }
+
+    .tab__inner.is-sticky .tab__settings {
+        max-width: 370px;
+        max-height: 100%;
+        overflow-y: auto;
+        position: fixed;
+        top: 0;
+        -webkit-overflow-scrolling: touch;
```
jamesjacobs: Minor thing, browser prefixed styles should go at the top of the block.

stylesheets/_layout.settings.scss
Stanton: There's a 1% margin on the tab settings that this bases a calculation on

jamesjacobs: 👍

PR #1101, opened 2019-12-04
Give panels a face lift

jackbentley:
Reduce panel title size
Add a separator bar between title and content
Increase panel default padding

**Before**
Screen Shot 2019-12-04 at 14 33 01

**After**
Screen Shot 2019-12-04 at 14 23 40

### DIFF
--- a/stylesheets/_component.panels.scss
+++ b/stylesheets/_component.panels.scss
@@ -7,18 +7,25 @@
     &__title {
         border: 0;
         clear: none;
-        font-size: $font-size-xlarge;
-        line-height: 32px;
+        font-size: $font-size-large * 1.1;
+        line-height: $font-size-large * 1.3;
         margin-bottom: 0;
-        padding: 8px 5px 0 10px;
+        padding: 12px $padding-large-horizontal;
+        background: rgba(0, 0, 0, .01);
+        border-bottom: 1px solid rgba(0, 0, 0, .1);
 
         i::before {
           width: 30px;
         }
+
+        .panel--padded & {
+            margin-top: -($padding-large-vertical * 2);
+            padding-top: 12px + ($padding-large-vertical * 2);
+        }
     }
 
     &__body {
-        padding: 10px $padding-base-horizontal;
+        padding: 10px $padding-large-horizontal;
 
         p,
         ul {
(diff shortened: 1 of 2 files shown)


stylesheets/_component.panels.scss
```diff
@@ -7,18 +7,25 @@
     &__title {
         border: 0;
         clear: none;
-        font-size: $font-size-xlarge;
-        line-height: 32px;
+        font-size: $font-size-large * 1.1;
+        line-height: $font-size-large * 1.3;
```
jamesjacobs: Line height needs to be 1.5 times the font-size to meet WCAG 1.4.12 Text Spacing (AA). May also need to tweak the heading icon position to ensure vertical alignment after increasing line-height.

stylesheets/_component.panels.scss
```diff
@@ -7,18 +7,25 @@
...
         margin-bottom: 0;
-        padding: 8px 5px 0 10px;
+        padding: 12px $padding-large-horizontal;
+        background: rgba(0, 0, 0, .01);
```
jamesjacobs: Properties within rules should be alpha ordered.

stylesheets/_component.panels.scss
```diff
@@ -108,8 +115,8 @@
 }
 
 .panel--padded {
-    padding-top: 24px;
-    padding-bottom: 24px;
+    padding-top: ($padding-large-vertical * 2);
```
jamesjacobs: Properties within rules should be alpha ordered.

stylesheets/_component.panels.scss
Stanton: Doesn't need to be, just has to not break if the line height is set to 1.5.

stylesheets/_component.panels.scss
jackbentley: Should probably add an automated test for this 😄

stylesheets/_component.panels.scss
jamesjacobs: Fixed

stylesheets/_component.panels.scss
jamesjacobs: Fixed

PR #1102, opened 2019-12-05
Repeater a11y fixes

### LINKED ISSUE #1071: Repeater component accessibility issues
The [repeater component](https://pulsar.docs.jadu.net/forms/repeater) contains various accessibility issues.

1. There are inputs without labels within `repeater__saved-data`. Could these be swapped for `type="hidden"`?
2. When the repeater form expands (either by add another or edit), focus should be set to the first input.
3. When the repeater form is closed, focus should be set to the "Add Another" button.
4. The "Add another" button is visually disabled when the form is open but can still be focussed. The `disabled` attribute should also be added.
5. The table should have a `caption` which matches the `label`.
6. There are various mark up validation failures relating to the table.

jamesjacobs:
This resolves various issues found in the repeater component, including:

1. Focus management for add, edit, save and close button actions
2. Changing add, save, cancel and edit buttons from links (`<a href="save">`) to `<button>`'s
3. Properly disabling the add button rather than just visually disabling it
4. Fixing issues where select2 child element IDs weren't unique after being cloned
5. Validation failures
6. Added a visually hidden caption to the repeater table
7. Adds actions column th text

Should be noted that SiteImprove still throws errors relating to hidden inputs not having a label. As their container has `display: none` this is not an issue and cannot be read by AT (confirmed in standard SRs).

Resolves #1071

### DIFF
--- a/js/FocusManagementService.js
+++ b/js/FocusManagementService.js
@@ -1,0 +1,42 @@
+const $ = require('jquery'); // might not need
+
+class FocusManagementService {
+    /**
+     * Focus Management Service
+     */
+    constructor () {
+        this.$element = null;
+        this.focusableElementList = 'a[href], input, select, textarea, button';
+    }
+
+    /**
+     * Store element that focus will be returned to
+     * @param element {HTMLElement}
+     */
+    storeElement ($element) {
+        this.$element = $element;
+    }
+
+    /**
+     * Return focus to original element
+     */
+    returnFocusToElement () {
+        this.$element.focus();
+    }
+
+    /**
+     * Move focus to first focuable element in a collection
+     * @param jquery? or do we want to jqery in here?
+     */
+    focusFirstFocusableElement ($collection) {
+        const $focusableElements = $collection
+            .find(this.focusableElementList)
+            .not('[tabindex=-1], [disabled], :hidden, [aria-hidden]');
+
+        if ($focusableElements.length) {
+            $focusableElements.first().focus();
+        }
+    }
+}
+
+module.exports = FocusManagementService;
--- a/js/FocusManagementService.js
+++ b/js/FocusManagementService.js
@@ -1,5 +1,3 @@
-const $ = require('jquery'); // might not need
-
 class FocusManagementService {
     /**
      * Focus Management Service
--- a/js/FocusManagementService.js
+++ b/js/FocusManagementService.js
@@ -9,7 +9,7 @@ class FocusManagementService {
 
     /**
      * Store element that focus will be returned to
-     * @param element {HTMLElement}
+     * @param {jQuery} $element - jQuery wrapper of the element
      */
     storeElement ($element) {
         this.$element = $element;
@@ -24,7 +24,7 @@ class FocusManagementService {
 
     /**
      * Move focus to first focuable element in a collection
-     * @param jquery? or do we want to jqery in here?
+     * @param {jQuery} $collection - jQuery collection of elements
      */
     focusFirstFocusableElement ($collection) {
         const $focusableElements = $collection
--- a/js/Repeater/RepeaterComponent.js
+++ b/js/Repeater/RepeaterComponent.js
@@ -341,9 +341,7 @@ class Repeater {
      * @param event
      */
     handleEditGroup (repeaterId, event) {
-        const edit = this.queryService.get('preview-root')
-            .querySelector(`[${this.queryService.getAttr('edit-id')}="${repeaterId}"]`);
-        const $edit = $(edit);
+        const $edit = $(this.queryService.get('preview-root').querySelector(`[${this.queryService.getAttr('edit-id')}="${repeaterId}"]`));
 
         event.preventDefault();
         this.repeaterPreviewService.toggleUi();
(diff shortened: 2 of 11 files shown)


js/FocusManagementService.js
```diff
@@ -0,0 +1,40 @@
...
+        this.focusableElementList = 'a[href], input, select, textarea, button';
+    }
+
+    /**
+     * Store element that focus will be returned to
+     * @param element {HTMLElement}
+     */
+    storeElement ($element) {
+        this.$element = $element;
+    }
```
mikedevelops: is `$element` here a `HTMLElement` or a jQuery collection? I suggest we reserve the dollar sign convention for jq objects only?

js/FocusManagementService.js
```diff
@@ -0,0 +1,40 @@
...
+    /**
+     * Return focus to original element
+     */
+    returnFocusToElement () {
+        this.$element.focus();
+    }
+
+    /**
+     * Move focus to first focuable element in a collection
+     * @param jquery? or do we want to jqery in here?
```
mikedevelops: Looks like you made your mind up, we should update this

js/Repeater/RepeaterComponent.js
```diff
@@ -325,11 +343,20 @@ class Repeater {
     handleEditGroup (repeaterId, event) {
         const edit = this.queryService.get('preview-root')
             .querySelector(`[${this.queryService.getAttr('edit-id')}="${repeaterId}"]`);
+        const $edit = $(edit);
```
mikedevelops: feels odd to have `$edit` & `edit` variables in the same scope here, should we just create the jq object and be done with it?

```
const $edit = $(this.queryService.get('preview-root')
    .querySelector(`[${this.queryService.getAttr('edit-id')}="${repeaterId}"]`));
```

mikedevelops: Also as an aside, I really regret creating the `queryService`, what on earth was I thinking 🙈

js/Repeater/RepeaterComponent.js
```diff
@@ -363,7 +389,9 @@ class Repeater {
 
         // Enable "add group" button if we have not exceeded max saved entries
         if (this.savedEntries < this.maxSavedGroups) {
-            $(this.queryService.get('add-group-button')).removeClass('disabled');
+            $(this.queryService.get('add-group-button'))
+            .removeClass('disabled')
+            .removeAttr('disabled');
```
mikedevelops: some funky indentation here, should we nudge these along

js/Repeater/RepeaterPreviewService.js
```diff
@@ -97,7 +97,14 @@ class RepeaterPreviewService {
                     true;
             })
             .forEach(preview => {
-                $(preview).find(this.queryService.getQuery('preview-ui')).toggleClass('disabled');
+                let $previewUi = $(preview).find(this.queryService.getQuery('preview-ui'));
+                $previewUi.toggleClass('disabled');
+
+                if ($previewUi.attr('disabled')) {
```
mikedevelops: Was just reading up on whether this is the best way to check, jq recommend we use `prop()` here 

> As of jQuery 1.6, the .attr() method returns undefined for attributes that have not been set. To retrieve and change DOM properties such as the checked, selected, or disabled state of form elements, use the .prop() method.

so maybe a more robust solution might be `if ($previewUi.prop('disabled') !== undefined) {`

js/utilities/UniqueIdService.js
```diff
@@ -40,6 +40,37 @@ class UniqueIdService {
...
+     * Update selectWoo child elements with unique IDs
+     * @param group {HTMLElement}
+     */
+    uniquifySelectWoo (group) {
+        const orginalSelect = group.querySelector('select.js-select2');
+        const orginalSelectId = orginalSelect.getAttribute('id');
+
+        [].slice.call(group.querySelectorAll('.select2-container'))
+            .forEach(select2Container => {
+                if (select2Container.querySelector('.select2-selection--multiple')) {
```
mikedevelops: Can we be explicit on these if statements in this method, not essential in this scenario, but it always feels a gamble to me to hand off the expression return value to the "truthy" / "falsy" gods

Seems more robust to stick to expressions that return boolean values (but it might just be me!)

```if (select2Container.querySelector('.select2-selection--multiple') !== null) {```

File: tests/js/web/FocusManagementServiceTest.js [hunk 7]
```diff
@@ -0,0 +1,119 @@
...
+
+    afterEach(() => {
+        $body.empty();
+    });
+
+    describe('storeElement()', () => {
+        it('should store the element', () => {
+            focusManagementService.storeElement($link);
+
+            expect(focusManagementService.$element === $link).to.be.true;
```
mikedevelops: ```expect(focusManagementService.$element).to.be($link)``` might make the error output easier to reason about if this should fail in the future

tests/js/web/FocusManagementServiceTest.js
```diff
@@ -0,0 +1,119 @@
...
+        it('should focus the first focusable element, if the element is a button ', () => {
+            $button.prependTo($body);
+
+            focusManagementService.focusFirstFocusableElement($body);
+
+            expect($button.is(':focus')).to.be.true;
+        });
+
+        it('should not focus the first element if it has tabindex="-1"', () => {
+            let $tabIndexZeroElement = $('<a href="#" tabindex="-1">Can not focus me</a>').prependTo($body);
```
mikedevelops: these _could_ be `const` regardless of how flakey the `const` implementation is!

js/FocusManagementService.js
jamesjacobs: It's a jQuery collection - I've updated the docblock

PR #1126, opened 2020-02-04
Hide add-filter button when all filters in a loaded filter have been used

### LINKED ISSUE #1116: Filter bar on load does not remove add button when all filters in use
**Describe the bug**
When rendering a filter bar on load with all the available filters being in use, the "Add" button still shows. The dropdown is empty.

This is different to when a page is loaded with no filters in use and you add all filters individually. The JS will remove the add button.

**To Reproduce**
Steps to reproduce the behaviour:
1. Load a page with a filter bar
2. Add all filters
3. Note the add button has disappeared
4. Submit the filter form. The page will reload.

**Expected behaviour**
The "Add" button remains disappeared (it does not show).

**Actual behaviour** 
The add button disappears after adding all of the available filters. The add button then re-appears after submitting the filter form. 

**Screenshots**
Screenshot 2020-01-16 at 14 43 53

 - OS: MacOS
 - Browser: Chrome
 - Version: 79.0.3945.117

Stanton:
This prevents situations where an empty list is displayed when clicking the add filter button.

filter

Closes #1116


js/FilterBarComponent.js
```diff
@@ -456,12 +457,20 @@ function populateFilterList ($filterbar) {
...
             // Hide the filter item in the list
             updateFilterList($addFilterButton, filterId, 'hide');
 
             // Show the filterbar
             $filterbar.removeClass('display--none');
         }
+
+        // Don't show add filter button if all filters have been loaded
+        if ($formGroups.length === hiddenFormGroups) {
+            $addFilterButton.addClass('display--none');
```
jamesjacobs: Should we not be using `u-display--none` here instead to try move towards the newer utility classes.

tests/js/web/FilterBarComponentTest.js
```diff
@@ -508,5 +508,9 @@ describe('FilterBarComponent', function () {
 		it('should add a label to the filterbar for select inputs', function () {
 			expect(this.$container.find('span[data-filter-id="size"]')).to.have.length(1);
 		});
+
+		it('should hide the add filter button if all filters have been used', function () {
```
jamesjacobs: Might be missing something but why is this passing? There's 4 inputs in the test DOM and only three of them have values when this test is run.

Stanton: There's an odd bug here where the colour select2 is being treated as if there's a value, this also happens when I take the test fieldset into a Lexicon UI to test...

PR #1129, opened 2020-02-05
fix issue where notifications dropdown wasn't visibly opening

Stanton:
The notifications event target ends up being the icon, and it doesn’t use a btn__group therefore the previous `closest(Selectors.BTN_GROUP)` never hit it and the notifications list was closed immediately after opening.

This change also checks the closest `.dropdown` class to look for the open class, which seems to make all the toolbar dropdowns work again.


Stanton:
Have also included this in the `flatcap` branch so it can be tested in current CMS

PR #1142, opened 2020-02-11
Restore ability to switch off datatable row selection with a data attribute

Stanton:
This had broken somewhat when the new centralised options ability was added.

This change also now removes the selection `td` elements entirely if the helper is used, which hasn't happened before.


views/pulsar/v2/helpers/html.html.twig
```diff
@@ -1577,14 +1577,16 @@ Datatable - https://jadu.github.io/pulsar/components/datatable/
...
-                            <input type="checkbox" class="form__control checkbox js-select-all" aria-label="Select all rows" data-toggle="tooltips" title="select all rows" data-placement="right" data-container="body" />
-                        </th>
-                    {% endif %}
-                    <th{{ attributes(column.attrs|default) }}>{{ column.title }}</th>
-                {% endfor %}
+                    {% for column in options.data.columns %}
+                        {% if loop.first %}
+                            {% if options['data-select'] is not defined or (options['data-select'] is defined and options['data-select'] != 'false') %}
+                                <th class="table-selection" data-orderable="false">
+                                    <input type="checkbox" class="form__control checkbox js-select-all" aria-label="Select all rows" data-toggle="tooltips" title="select all rows" data-placement="right" data-container="body" />
```
jamesjacobs: This needs updating to use the new version of tooltips.

Stanton:
> Tooltips need updating to new version in DT help changes

Done

PR #1148, opened 2020-02-24
Repeater bugfixes

Stanton:
### Fix issue where add button is not disabled when the max-items limit has been exceeded.
The way the repeater loops through each previously saved item means if there’s more than one item the add button will always be enabled, but never re-disabled if the number of items loaded meets or exceeds the `max-items` option.


Stanton:
I completely botched this PR up with resolving conflicts, creating new PR now

PR #1149, opened 2020-02-26
Accessible tooltips

### LINKED ISSUE #1019: Tooltips aren't accessible
**Describe the bug**
Existing tooltips don't offer the same experience to SR users as the tooltip contents aren't read when the triggering element is focussed/hovered. The only way for a SR user to find/read the tooltip contents is to focus the trigger and use the SR to navigate text on the page. Additionally they do not currently meet WCAG 1.4.13 at AA.

**Ideal behaviour**
There's a number of things to consider here:

1. The tooltip contents should be read when the triggering element is focussed. Depending on the use case this could be with either `aria-labbelledby` (in the case of a button with a count bubble) or `aria-describedby` for general tooltip usage.

2. To meet 1.4.13 (Content on Hover or Focus AA), the tooltip must be:
 **Dismissable** -  without moving the pointer or KB focus. So we need to be able to close the tooltip with ESC.
**Hoverable**  - the tooltip contents need to stay visible whilst the are hovered over 
**Persistent** - the tooltip content remains visible until the hover or focus trigger is removed, the user dismisses it or the information is no longer valid.

3. As popovers are powered by the existing tooltip.js, for backwards compatibility it's worth adding new tooltip functionality to a new JS module with a different `data` attribute hook.

**Proposed solution**
Example outputted markup:
```
<button aria-describedby="guid-1234">  
  Example button with tooltip
</button>  
<div role="tooltip" id="guid-1234">Tooltip contents</div>  
```

The `aria-describedby` and `role="tooltip"` `div` could be added via JS using a data attribute for both the selector and contents. For example:

```
<button data-tooltip="Tooltip contents">  
  Example button with tooltip
</button> 
```

We'd need to ensure there's no clash between existing `tooltip.js` functionality and new functionality. Potentially by removing the `tooltip.js` markup / unbinding any events/listeners in case both new and old attributes are used. We could also remove any `title` attribute on the triggering element too as they don't offer much in the way of accessibility.

It may be necessary to wrap the triggering element and the added tooltip `div` for positioning reasons.

jamesjacobs:
This change adds a new WCAG 2.1 compliant Tooltip implementation. It uses the [Tippy.js](https://atomiks.github.io/tippyjs/) library with some custom configuration to address missing WCAG requirements. Existing tooltips are untouched.

### New behaviour
- Tooltip content can be hovered
- Tooltip content is announced by screen reader when the triggering element is focussed
- Tooltips can be closed by pressing ESC
- By default, Tooltips display above the triggering element (can be changed via data attributes), though they will appear elsewhere (right, bottom, left) if there isn't space available (for example, if the element is at the top of the viewport or if there isn't space available on mobile).  

These new tooltips cover the requirements of [WCAG 2.1 1.4.13.](https://www.w3.org/WAI/WCAG21/Understanding/content-on-hover-or-focus.html) (Dismissable, Hoverable and Persistent)

### Testing

Browser tested in:
- Chrome
- Safari
- Firefox
- Edge
- IE11
- IE10
- iOS

Screen reader tested in:

- Safari + Mac VO
- FF + NVDA
- IE11 + JAWS
- IE11 + NVDA
- Chrome + NVDA
- iOS + VO

## Breaking change checklist

| Does this pull request... | ⚠️ Yes / ✅ No |
| ---- | ---- |
| require changes to existing unit tests |  |

Documentation changes to follow.

Fixes #1019

### DIFF
--- a/js/Tooltips/TooltipListener.js
+++ b/js/Tooltips/TooltipListener.js
@@ -39,7 +39,7 @@ class TooltipListener {
             // Interactive allows us the hover over the tooltip contents, needed for WCAG 1.4.13: Content on Hover or Focus
             // unfortunately, adding aria: 'describedby' here doesn't work as the interactive option removes it and replaces
             // with aria-expanded see https://github.com/atomiks/tippyjs/issues/709
-            // therefore we manually handle the toltip aria-describedby attribute in the onMount and onHide lifecycle hooks
+            // therefore we manually handle the tooltip aria-describedby attribute in the onMount and onHide lifecycle hooks
             interactive: true,
 
             // Custom methods, also required for WCAG compliance
--- a/js/Tooltips/TooltipListener.js
+++ b/js/Tooltips/TooltipListener.js
@@ -1,0 +1,100 @@
+'use strict';
+
+/**
+ * Accessible tooltips using the tippyjs library with some custom behaviour to meet WCAG 2.1 1.4.13
+ */
+class TooltipListener {
+
+    /**
+     * Tooltip listener
+     * @constructor
+     * @param {jQuery} $html - jQuery wrapper of the html node
+     * @param {tippy} TippyJS lib
+     * @param {hideAll} Tippy hideAll method
+     */
+    constructor ($html, tippy, hideAll) {
+        this.$html = $html;
+        this.tippy = tippy;
+        this.hideAll = hideAll
+    }
+
+    /**
+     * Initialise
+     */
+    init () {
+        this.tippy('[data-tippy-content]', {
+
+            // Default to not allowing html inside of tooltip
+            allowHTML: false,
+
+            // Remove the default fade animation
+            animation: 'none',
+
+            // To prevent overflow issues, always append to body
+            appendTo: this.$html.find('body')[0],
+
+            // Remove the animation duration
+            duration: 0,
+
+            // Interactive allows us the hover over the tooltip contents, needed for WCAG 1.4.13: Content on Hover or Focus
+            // unfortunately, adding aria: 'describedby' here doesn't work as the interactive option removes it and replaces
+            // with aria-expanded see https://github.com/atomiks/tippyjs/issues/709
+            // therefore we manually handle the toltip aria-describedby attribute in the onMount and onHide lifecycle hooks
+            interactive: true,
+
+            // Custom methods, also required for WCAG compliance
+            onCreate: this.onCreate,
+            onMount: this.onMount,
+            onHide: this.onHide,
+            onHidden: this.onHidden
+        });
+
+        // Close ESC button
+        this.$html.on('keydown', (event) => {
+            if (event.keyCode === 27) {
+                this.hideAll();
+            }
+        });
+    }
+
+    onCreate (instance) {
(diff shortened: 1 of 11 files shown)


js/Tooltips/TooltipListener.js
```diff
@@ -0,0 +1,100 @@
...
+            // To prevent overflow issues, always append to body
+            appendTo: this.$html.find('body')[0],
+
+            // Remove the animation duration
+            duration: 0,
+
+            // Interactive allows us the hover over the tooltip contents, needed for WCAG 1.4.13: Content on Hover or Focus
+            // unfortunately, adding aria: 'describedby' here doesn't work as the interactive option removes it and replaces
+            // with aria-expanded see https://github.com/atomiks/tippyjs/issues/709
+            // therefore we manually handle the toltip aria-describedby attribute in the onMount and onHide lifecycle hooks
```
Stanton: toltip

Stanton:
Added breaking change checklist to be completed

Stanton:
Are the select2 and toggle examples supposed to work?

tips

jamesjacobs:
@Stanton select2's and toggles created using the helper don't currently support tooltips (existing issue). Was going to investigate the toggle issue in a separate ticket, though I'm not sure it's going to be possible at all on select2s.

js/Tooltips/tooltipsFactory.js
```diff
@@ -0,0 +1,17 @@
...
+const TooltipListener = require('./TooltipListener');
+
+/**
+ * Create a Tooltip listener instance
+ * @param {jQuery} $html - jQuery wrapper of the html node
+ */
+function tooltipFactory ($html) {
+    const tooltipListener = new TooltipListener($html, tippy, hideAll);
+
+    tooltipListener.init();
```
mikedevelops: let's remove this init here as we'll do this where we init everything else. We want to keep our factories declarative and side-effect free.

we want to return the `tooltipListener` instance from this function

js/main.js
```diff
@@ -53,6 +53,7 @@
         pulsar.tableDetail.init();
         pulsar.dropZoneComponent = pulsar.DropZoneComponentFactory.create($('body')[0], '.dropzone');
         pulsar.repeaterManager.init();
+        pulsar.tooltipFactory($html);
```
mikedevelops: so here we just need to call our factory method to get our instance and then call init on the instance

mikedevelops: so this might look something like...

```
// We'll put the tooltip listener on pulsar's global object here...
pulsar.tooltipListener = pulsar.tooltipFactory($html);

// Then invoke it's init method
pulsar.tooltipListener.init();
```
```

File: tests/js/web/Tooltips/TooltipListenerTest.js [hunk 4]
```diff
@@ -0,0 +1,134 @@
...
+    let $body;
+    let tooltipListener;
+    let $button;
+    let tippyStub;
+    let hideAllStub;
+    let keyDownEvent = $.Event('keydown');
+
+    beforeEach(() => {
+        $html = $('html');
+        $body = $('body');
```
mikedevelops: can we use fake html and body nodes here? Like just regular divs? This sort of stuff will break the browser version of the test suite (which are admittedly broken already!)

Issue #1165, opened 2020-03-18
Alt attribute should always be present (html.media)


jamesjacobs:
Mark up generated by the `html.media` helper only include the `alt` attribute if a value was passed via the `image_alt` option. The `alt` attribute should always be present on images. If no text option was used, it should default to empty `alt=""`.

See the following error from Site Improve on the lexicon page:
![image](https://user-images.githubusercontent.com/756393/76982043-3adc8300-6933-11ea-9863-d1fac4a6e64e.png)

jamesjacobs:
Fixed by #1188

PR #1188, opened 2020-03-20
Make media component always have alt attribute

Stanton:
If no alt text supplied the helper will render with an empty `alt` attribute.

**Given**
```
{{
    html.media({
        'image': 'foo.png',
        'title': 'Bar',
        'description': 'Baz'
    })
}}
```

**Before**
```
<div class="media">
    <img src="foo.png" class="media__image" />
    <div class="media__body">
        <p>Bar</p>
        <span class="small-type muted">Baz</span>
    </div>
</div>
```

**After**
```
<div class="media">
    <img src="foo.png" alt="" class="media__image" />
    <div class="media__body">
        <p>Bar</p>
        <span class="small-type muted">Baz</span>
    </div>
</div>
```
This PR also updates the phpunit suite to work with phpunit 9.


jamesjacobs:
Fixes #1165

Stanton:
Codecov looks like another false positive

PR #1190, opened 2020-03-23
Fix range focus site improve error

### LINKED ISSUE #1173: form.range focus issues
Site improve reports focus errors on range inputs. Currently the "thumb" is highlighted on focus but SI expects the entire input to be highlighted. Investigate.

image

jamesjacobs:
This change adds a visual focus indicator to the range input in addition to the existing thumb focus style.

After change:

image

Fixes #1173

### DIFF
--- a/stylesheets/_mixin.focus.scss
+++ b/stylesheets/_mixin.focus.scss
@@ -53,8 +53,11 @@
 
 @mixin pulsar-range-focused {
     border: 0;
-    box-shadow: none;
-    outline: none;
+    border-radius: 0;
+    box-shadow: inset 0 0 0 2px color(black);
+    outline: 3px solid color(border, focus);
+    outline-offset: 0;
+    padding: 0 5px;
 
     &::-webkit-slider-thumb {
         box-shadow: 0 0 0 3px color(white), 0 0 0 6px color(black), 0 0 0 9px color(border, focus);


stylesheets/_mixin.focus.scss
```diff
@@ -53,8 +53,11 @@
 
 @mixin pulsar-range-focused {
     border: 0;
-    box-shadow: none;
-    outline: none;
+    border-radius: 0;
```
Stanton: Perhaps it would make sense to remove this higher up, seeing as it's an inert condition for range sliders?

stylesheets/_mixin.focus.scss
```diff
@@ -53,8 +53,11 @@
...
+    box-shadow: inset 0 0 0 2px color(black);
+    outline: 3px solid color(border, focus);
+    outline-offset: 0;
+    padding: 0 5px;
```
Stanton: Does this cause a jump on focus?

jamesjacobs: Not a jump as such but the range width does shrink a little on focus but with the focus styling it's barely perceivable. Without the padding the outline almost overlaid the range track.

jamesjacobs: range focus

Stanton: Ah that's fine then, I was worried about a vertical jump 👍

stylesheets/_mixin.focus.scss
jamesjacobs: Needed to override default browser focus styling in a couple of browsers.

PR #1192, opened 2020-03-23
Add aria-disabled attribute to disabled UI wrapper and change text color for disabled links

### LINKED ISSUE #1177: Disabled UI a11y issues
Legends, text and link within a "disabled UI" pattern fail colour contrast. Potentially just needs aria-disabled="true" adding to the `fieldset` (or `form`?). Also, links within the disabled ui can still be tabbed to / gain focus.

image

Stanton:
Using opacity instead of a lighter `color` achieves the same visual effect, keeps a11y tools happy, and meets WCAG exceptions for disabled element contrast.

Also uses opacity for disabled choice labels

**Before**
Screenshot 2020-03-24 at 09 00 53
Using `color: color(text, disabled);` (`#bebebe`)

**After**
Screenshot 2020-03-24 at 09 00 39

This uses `opacity: .4;`, a touch darker than before, but still enough, I feel.

Closes #1177

### DIFF
--- a/js/DisableUiComponent.js
+++ b/js/DisableUiComponent.js
@@ -40,7 +40,7 @@ DisableUiComponent.prototype.disable = function (target) {
             .addClass('js-disabled u-cursor-not-allowed');
 
         // Wrap with disabled wrapper to visually disable
-        $this.wrap('<div class="u-ui-disabled"></div>');
+        $this.wrap('<div class="u-ui-disabled" aria-disabled="true"></div>');
     });
 };
 
@@ -53,7 +53,7 @@ DisableUiComponent.prototype.enable = function (target) {
         var $this = $(this);
 
         // Remove attribute used to disable a UI on load
-        $this.removeAttr('data-disable-ui');
+        $this.removeAttr('data-disable-ui, aria-disabled');
 
         // Enable form elements
         $this.find(FORM_ELEMENTS)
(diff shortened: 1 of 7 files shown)


js/DisableUiComponent.js
```diff
@@ -37,10 +37,11 @@ DisableUiComponent.prototype.disable = function (target) {
         // Disable links (uses .js-disable as any existing disabled)
         $this.find(LINK_ELEMENTS)
             .on('click', preventDefaultAndStopPropagation)
-            .addClass('js-disabled u-cursor-not-allowed');
+            .addClass('js-disabled u-cursor-not-allowed')
+            .prop('aria-disabled', 'true');
```
jamesjacobs: Minor point, we use a mix of `attr` and `prop` here. I'd stick to `attr` unless we are specifically changing a `prop`

js/DisableUiComponent.js
```diff
@@ -67,7 +68,8 @@ DisableUiComponent.prototype.enable = function (target) {
         // Enable links
         $this.find(LINK_ELEMENTS)
             .off('click', preventDefaultAndStopPropagation)
-            .removeClass('js-disabled u-cursor-not-allowed');
+            .removeClass('js-disabled u-cursor-not-allowed')
+            .removeProp('aria-disabled');
```
jamesjacobs: `attr` vs `prop`?

PR #1195, opened 2020-03-25
Add accessible label for user guides toolbar button

Stanton:


views/pulsar/v2/helpers/nav.html.twig
```diff
@@ -172,7 +172,7 @@
         {{
             html.button({
                 'class': 'dropdown-toggle btn--outline btn--inverse',
-                'label': html.icon('book') ~ ' <span class="caret"></span>',
+                'label': html.icon('book', { 'label': 'User guides' }) ~ ' <span class="caret"></span>',
```
jamesjacobs: Should this be an option so we can pass trans when needed?

Stanton: We have little to no existing provision for this, or common guidelines for how to provide translation provisions in Pulsar. Happy to do so if you can give an example of how you would use it in CXM.

PR #1211, opened 2020-04-15
Restrict toggle activation to the control label and the toggle control only

### LINKED ISSUE #927: Toggle switch can be toggled in unexpected places
**Describe the bug**
Toggle buttons can be toggled in odd places. Looks as though a combination of toggles being inside labels and labels being full width means I can click wayyy off to the right and toggle a toggle by mistake.

**To Reproduce**
This can be reproduced within CXM in Settings -> System -> Integrations HUB.
I suspect this can be reproduced in the Lexicon also.

**Expected behavior**
I expect to click the label and toggle the toggle, however it's unexpected that I can click to the right of the toggle (which is actually label) and toggle the toggle.

**Screenshots**
toggle

 - OS: OSX High Sierra
 - Browser: Chrome
 - Version: 70

### LINKED ISSUE #1232: Clicking a toggle too fast causes focus style to be lost
**Describe the bug**
Clicking a toggle button too fast causes the focus style to be lost. **The toggle is still in focus, however.**

**To Reproduce**
Steps to reproduce the behaviour:
1. Click on a toggle switch fast

**Expected behaviour**
Focus style stays in place.

**Screenshots**
toggle-focus

Stanton:
* Prevents clicks on the large surrounding label, including help text, from toggling the control
* Trigger the focus event on the hidden input when clicked

**Before**
toggle-before
**After**
toggle-after

Closes #927
Closes #1232


Stanton:
@jamesjacobs IE11 issue is related to an issue with timepicker which breaks the bundle, should be evident through IE11 console if you need to verify https://github.com/jonthornton/jquery-timepicker/issues/710

Updating this branch to timepicker 1.13.3 allows this change to be properly tested.

PR #1221, opened 2020-05-01
Add button styling to file upload choose file button in webkit

jackbentley:
**Before:**
Screen Shot 2020-05-01 at 19 03 25

**After:**
Screen Shot 2020-05-01 at 19 02 39

### DIFF
--- a/stylesheets/_component.buttons.scss
+++ b/stylesheets/_component.buttons.scss
@@ -40,7 +40,8 @@ $btn-small-padding-x:       $btn-padding-x / 2.5;
 $btn-white:                 'white';
 // scss-lint:enable StringQuotes
 
-.btn {
+.btn,
+::-webkit-file-upload-button {
     @include inline-block;
     -webkit-appearance: none; /* 1 */
     background: transparent;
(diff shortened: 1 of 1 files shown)


jackbentley:
Might be worth adding `::-ms-browse` as well?

https://developer.mozilla.org/en-US/docs/Web/CSS/::-ms-browse

jackbentley:
Yeah, the button only receives focus styles when it's directly clicked/active. However, tabbing to the input will add the same focus as it does currently. It's more of an eye-candy thing really.

Click button:
Screen Shot 2020-05-19 at 21 30 24

Click label or area that isn't the button or tab:
Screen Shot 2020-05-19 at 21 30 05

You could make it always only show button focus by altering the focus styles for the input to be something like `input[type="file"]:focus ::-webkit-file-upload-button` instead

Alternatively, we could take the approach of hiding the native file input and wrapping it in a label with a button. Could even use JS to show the current file - https://stackoverflow.com/a/47974444/1600814

The `webkit` prefix for this should work in Chrome, Safari and Edge. The `ms` prefix will work in IE. Firefox doesn't seem to have an equivalent.

Stanton:
Visually this is fine, however as mentioned it needs refactoring as mentioned to reduce the compiled CSS footprint.

Blocking for now, will close if no further movement.

PR #1231, opened 2020-05-20
Enforce the use of IDs in form helpers

### LINKED ISSUE #795: Enforce the use of IDs within form helpers
`id` attributes are currently optional within form helpers which—while perfectly valid HTML—means that the markup generated isn't compliant with WCAG 2.0 by default and will throw a 'missing label' error when using tools such as WAVE or SiteImprove. The current assumption that developers will always supply an `id` to form fields cannot be relied upon to assure compliance.

### Example
*Input:*
```
{{
    form.text({
        'label': 'Text'
    })
))
```
*Output:*
```
<div class="form__group">
    <label class="control__label">Text</label>
    <div class="controls">
        <input type="text" class="form__control">
    </div>
</div>
```

If an `id` is supplied, the label will be correctly updated to refer to the input, but this relies completely on developers supplying an `id` to all instances of a helper.

### Example
*Input:*
```
{{
    form.text({
        'label': 'Text',
        'id': 'myInput'
    })
))
```
*Output:*
```
<div class="form__group">
    <label for="myInput" class="control__label">Text</label>
    <div class="controls">
        <input id="myInput" type="text" class="form__control">
    </div>
</div>
```

# Proposal

To ensure Pulsar form helpers meet current WCAG guidance as a default option:

* If no `id` is supplied, generate a guid and use it as the `id`
* Allow this behaviour to be overridden by supplying `'id': null`

### Example
*Input:*
```
{{
    form.text({
        'label': 'Text'
    })
))
```
*Output:*
```
<div class="form__group">
    <label for="guid-xxx" class="control__label">Text</label>
    <div class="controls">
        <input id="guid-xxx" type="text" class="form__control">
    </div>
</div>
```

Stanton:
**Second PR as there was some base branch shenanigans which complicated the previous PR**

Currently, all options passed to a form helper are optional, this places the main burden on developers to ensure they're passing the appropriate attributes based on their use case, however in the context of form helpers the omission of an `id` attribute, with the inclusion of a `label` cause the helper to create accessibility issues (in the form of orphaned labels due to the lack of a `for` attribute).

This PR makes form helpers generate unique `id` attributes if no custom `id` has been provided, allowing the form component to appropriately link up the label elements as required.

**Example:**
```html
<div class="form__group">
  <label for="id-guid-12345" class="control__label">Label</label>
  <div class="controls">
    <input type="text" id="id-guid-12345" aria-describedby="help-guid-12345" class="form__control" />
    <span class="help-block" id="help-guid-12345">Help text</span>
  </div>
</div>
```

GUIDs were previously being generated on help text items so they can be associated with an input through the `aria-describedby` attribute, for readability sake the format of the existing GUIDs have been changed slightly from `guid-{{ rand() }}` to `help-guid-{{ rand() }}`, and the new id GUID uses `id-guid-{{ rand() }}`.

**Important note about test fixtures**
In our fixtures the random number of all GUIDs are rewritten as `id-guid-1` and `help-guid-1` so that they can be tested easily, this may cause some confusion for 'compound' field types like form.choice, form.compound and form.button_group where there are multiple inputs all using the same ID, which may lead you to think this would generate invalid HTML, however in production, the GUID pairs (`id` and `for`) are appropriately unique for each set of inputs in a compound helper.

**Example**
```html
<fieldset id="id-guid-1870134239" class="form__group form-choice">
  <legend class="control__label">Choice radios</legend>
  <div class="controls">

      <label for="id-guid-138535425" class="control__label">
          <input value="one" name="one" id="id-guid-138535425" aria-describedby="help-guid-1269151066" type="radio" class="form__control radio">
          <span class="form-choice__label">One</span>
      </label>

      <label for="id-guid-574823111" class="control__label">
          <input value="two" name="one" id="id-guid-574823111" aria-describedby="help-guid-1269151066" type="radio" class="form__control radio">
          <span class="form-choice__label">Two</span>
      </label>

      <label for="id-guid-161742440" class="control__label">
          <input value="three" name="one" id="id-guid-161742440" aria-describedby="help-guid-1269151066" type="radio" class="form__control radio">
          <span class="form-choice__label">Three</span>
      </label>

      <span class="help-block" id="help-guid-1269151066">Help text to give more information about expected input</span>
  </div>
</fieldset>
```

Closes #795


Stanton:
Good candidate for 12.x release @jamesjacobs

jamesjacobs (changes requested):
Tested this locally (updated with develop) and found the following issues:
1. it sets a guid ID on the `fieldset` and `legend` on form.compounds. I don't see why this would cause an issue but raising it for awareness.
2. In form.compound, when an individual field is not passed an `ID` it throws a twig error. You can replicate this in the lexicon, by remove an `ID` from one of the DOB example fields, such as day.
3. Does not add the ID to toggle switches if no ID is passed in twig. Although the label wraps the input, we should still cover this for screen reader coverage.
4. There are some failing tests.

PR #1235, opened 2020-05-22
Removable label enhancements

jamesjacobs:
Existing removable labels (labels with a x icon or more recently an x button) had poor focus styles and some styling issues on focus. This change modifies the label mark up slightly to allow for clearer focus styling.

**Before (unfocussed):**

Screenshot 2020-05-22 at 10 39 00

**Before (first 2 focussed):**

Screenshot 2020-05-22 at 10 39 22

**After (unfocussed):**

Screenshot 2020-05-22 at 10 36 02

**After (first 2 focussed):**

Screenshot 2020-05-22 at 10 40 25

**After (filter bar example):**

Screenshot 2020-05-22 at 10 36 51

Existing remove button styles have been kept for backward compatibility. Due to some super specific CSS rules, I've had to use `!important` in a couple of places. I've also discovered an issue where labels styles are been added to the compiled CSS 3 times, which I'll address in a different ticket (and possibly some of the css specificity issues).

Docs PR to follow.


stylesheets/_component.labels.scss
```diff
@@ -45,12 +45,48 @@
...
+
+    .label__text {
+        padding: .25em .6em;
+        display: inline-block;
+    }
+
+    .btn.remove-button {
+        padding: .25em .45em;
+        background-color: rgba(0, 0, 0, .15);
+        border-radius: 0 4px 4px 0;
```
Stanton, line 60: `$border-radius`

jamesjacobs: Good spot, pushed a fix.

jackbentley:
There’s a similar background colour thing going on in the panel changes I made. Would be nice to check if they’re using the same opacity in the background rgba to keep consistency

PR #1238, opened 2020-05-27
Repeater focus management

jamesjacobs:
This change resolves the following:

**Focus management**

Before this change there were 3 scenarios where focus could be lost.

1. Save clicked and max entry limit reached. Focus now moves to the edit button of the newly created row.
2. Save clicked and max entry limit not reached. Focus now moves to Add Another button.
3. Delete clicked. Focus now moves to the delete button on the previous row. If no previous row is present, focus moves to Add button.

**Site Improve**

Site Improving incorrectly reported missing labels on hidden (Saved data) inputs. This was a false positive as inputs where hidden via `display: none` (which takes them out of the a11y tree). In order to silence SI and prevent future issues being raised, the associated input label text is now added to the cloned saved data input as an `aria-label`.

This branch also contains the changes made in #1201


js/FocusManagementService.js
```diff
@@ -22,6 +22,17 @@ class FocusManagementService {
...
+    /**
+     * Check if an element is stored
+     */
+    hasStoredElement () {
+        if (this.$element == null) {
+            return false;
+        } else {
+            return true;
+        }
+    }
```
mikedevelops, line 34: `return this.$element !== null` perhaps?

PR #1243, opened 2020-06-01
Start changing forms to use 100 width when used in modals/repeaters

Stanton:


Stanton:
Fixed merge conflict, needs another 👍 please @jamesjacobs

jamesjacobs (changes requested):
Feels like we should be cautious with this and perhaps keep it to it's own release after testing:
- full page modals
- form compound
- sanity check outputted CSS
- legacy form group modifiers

Also any potential impact on legacy modals.

Should also smoke test various modals in product prior to release.

PR #1249, opened 2020-06-02
Ensure interactive elements in closed mobile sidebar cannot be focussed or seen by SRs

### LINKED ISSUE #1247: Sidebar use of aria-hidden
Help sidebar uses `aria-hidden="true"` on mobile when the sidebar is closed. This will remove the element and its children from the accessibility tree. AXE flags this as an issue when the sidebar contains focusable elements such as a button or link.

The attribute is removed when the sidebar is opened so I'm not 100% convinced this is an issue however we should investigate and potentially change to prevent QA teams flagging it as an issue.

AXE issue link: https://dequeuniversity.com/rules/axe/3.5/aria-hidden-focus?application=AxeChrome

**Update**
It seems you can focus/tab to interactive elements within the closed help sidebar

jamesjacobs:
This fixes an issue where links/buttons within the mobile help sidebar could be tabbed to and be seen/navigated to by screen readers.

Fixes #1247

### DIFF
--- a/js/HelpTextComponent.js
+++ b/js/HelpTextComponent.js
@@ -176,7 +176,7 @@ HelpTextComponent.prototype.toggleChildElementInteractivity = function ($contain
     if (interactive === true) {
         $container.find(focusableElements).removeAttr('tabindex');
     } else {
-        $container.find(focusableElements).attr('tabindex', "-1");
+        $container.find(focusableElements).attr('tabindex', '-1');
     }
 }
 
--- a/js/HelpTextComponent.js
+++ b/js/HelpTextComponent.js
@@ -34,6 +34,9 @@ HelpTextComponent.prototype.init = function () {
         component.toggleHelpSidebar();
     });
 
+    // Make mobile help container child element non-interactive
+    component.toggleChildElementInteractivity($tabHelpContainer, false);
+
     // Bind to tab.js event to update help text in sidebar on tab change
     component.$html.find('a[data-toggle="tab"]').on('shown.bs.tab', function (e) {
         component.updateHelpSidebar();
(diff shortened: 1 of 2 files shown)


js/HelpTextComponent.js
```diff
@@ -156,4 +167,17 @@ HelpTextComponent.prototype.handleFocusOut = function (e) {
...
+/**
+ * Toggle focusability of elements
+ */
+HelpTextComponent.prototype.toggleChildElementInteractivity = function ($container, interactive) {
+    var focusableElements = 'a[href], button';
+
+    if (interactive === true) {
+        $container.find(focusableElements).removeAttr('tabindex');
+    } else {
+        $container.find(focusableElements).attr('tabindex', "-1");
```
Stanton, line 179: Mix of quote marks here

js/HelpTextComponent.js
```diff
@@ -156,4 +167,17 @@ HelpTextComponent.prototype.handleFocusOut = function (e) {
     }
 }
 
+/**
+ * Toggle focusability of elements
+ */
+HelpTextComponent.prototype.toggleChildElementInteractivity = function ($container, interactive) {
+    var focusableElements = 'a[href], button';
+
+    if (interactive === true) {
```
Stanton, line 176: should work with just `if (interactive) {` ?

js/HelpTextComponent.js
```diff
@@ -156,4 +167,17 @@ HelpTextComponent.prototype.handleFocusOut = function (e) {
     }
 }
 
+/**
+ * Toggle focusability of elements
+ */
+HelpTextComponent.prototype.toggleChildElementInteractivity = function ($container, interactive) {
+    var focusableElements = 'a[href], button';
```
Stanton, line 174: We've got two lists of elements now, might be useful to have this as some form of high-level config var? 

https://github.com/jadu/pulsar/blob/develop/js/TableDetailComponent.js#L10

js/HelpTextComponent.js
jamesjacobs: In this instance yeah, it would. I prefer to be more explicit though as it's possible the value of that var might change in the future.

js/HelpTextComponent.js
jamesjacobs: Possibly yeah, I think I've got a similar situation in the next issue I'm going to pick up related to the main nav so can investigate there.

PR #1258, opened 2020-06-10
Make tab.js responsible for managing <main> element placement

Stanton:
This change implements the following:

1. Stop twig creating `<main>` elements, in favour of divs
2. On page load, tab.js converts tab content in active tab pane to `<main>`, and adds skip target ID
3. On tab show event, new tab gets converted to `<main>`, hidden tab back to `<div>`

This should stop a11y tools reporting issues with multiple `<main>` landmarks and duplicate instances of `id="skip-target"`.


Stanton:
@jamesjacobs Tied into tab methods so non-js tabs will have no change

Stanton:
Cherry-picked into flatcap-july too

PR #1261, opened 2020-06-15
Remove/Edit icon button focus styles

jamesjacobs:

1. remove and edit icon buttons now have correct focus styles
2. help-block text within tfoot forms now meet AA contrast
3. added `component.edit-button` sass partial to main `pulsar.scss` file
4. updated lexicon remove/edit icon button examples
5. removed redundant datagrid lexicon page


views/lexicon/tables/form-tables.html.twig
```diff
@@ -123,7 +123,7 @@
                         <td>my_label</td>
                         <td>my_value</td>
                         <td class="centered">
-                            {{ html.edit_button({ 'placement': 'left' }) }}
+                            {{ html.edit_button({ 'placement': 'left', 'class': 'u-margin-right-small' }) }}
```
Stanton, line 126: Will still pass this but it feels like a button group may be more appropriate here to manage spacing?

PR #1263, opened 2020-06-16
JStree a11y review fixes

jamesjacobs:
This branch contains a slight change to the JS-tree webkit scroll bars to make them more consistent with over styled scrollbars.

**A11y review notes:**

JStree follows the aria-tree pattern and includes the appropriate keyboard navigation support. 2 potential issues where found (reported by Site Improve):

1. Element not highlighted on focus. This relates to the JS tree wrapper with tabindex="0". Adding focus styles does not resolve is issue as when focus shifts to the container, focus is immediately moved to the first element in the tree. Therefore, this seems to be a non-issue.
2. WAI-ARIA group is missing a name. This relates to the plugin using aria-activedescendant in combination with role="group" and a missing group name. There doesn't seem to be a way to resolve this.

Both of the above points will be added to the JStree entry in the accessible components documentation list. Screen reader support for aria-tree's is limited at best so this component should be avoided where possible.


stylesheets/_component.jstree.scss
```diff
@@ -12,23 +12,23 @@
...
+    &::-webkit-scrollbar {
+        height: 7px;
+        width: 7px;
+    }
 
-        &::-webkit-scrollbar {
-            width: 7px;
-        }
+    &::-webkit-scrollbar-thumb {
+        border-radius: 4px;
```
Stanton, line 25: $border-radius

jamesjacobs: Done

PR #1267, opened 2020-06-19
Remove unnecessary twig extensions

jackbentley:
These twig extensions are unnecessary and can be achieved with built-in Twig functionality. None of them is used in the Pulsar repo it's self. 

## `constant_defined`

Check if a constant is defined.

This can be replaced with the following:

```twig
{% if constant('MY_CONST') is defined %}
```

See https://twig.symfony.com/doc/1.x/functions/constant.html for more info

## `get_constant`

Output a constant if it is defined (with a fallback).

This can be replaced with the following:

```twig
{{ constant('MY_CONST') is defined ? constant('MY_CONST') : 'fallback' }}
```

Use of constants in these cases should be discouraged. A constant should be always defined. If it is not, then it's likely it shouldn't be a constant.

See https://twig.symfony.com/doc/1.x/functions/constant.html for more info


Stanton:
CMS and XFP both currently use `constant_defined` in production, we will not consider removing them until this is no longer the case. Recommend raising merge requests in the respective products and re-opening this when it would no longer represent a breaking change.

PR #1274, opened 2020-06-30
Stop flex toolbar sliding underneath main navigation in IE11

Stanton:
Fixes an issue introduced in the new flex box based toolbar for Flatcap branch.

**Before**
yay

**After**
yay2


Stanton:
Pretty sure we don't need this PR anymore as it's flatcap-july > flatcap.... Won't delete the branch at this point though.

Issue #1180, opened 2020-03-19
Rules pattern a11y issues


jamesjacobs:
### Issue 1
Axe complains about heading order in rules heading (currently `h5`).

### Issue 2
Unable to navigate with a screen reader to the form in the `tfoot`. Only tested on Mac VO at time of writing.

Stanton:
**Issue 1**
Doesn't feel appropriate to replace these with H1 elements, the general rule block sections are wrapped in fieldsets with a hidden legend (which interestingly doesn't seem to be properly announced with select2s, other fields are fine) so perhaps replace the `h5` with regular old `span` elements...

**Issue 2**
I seem to be able to navigate to this without issue, is there something I'm missing?

PR #1277, opened 2020-07-10
Rules UI a11y updates

jamesjacobs:

### Rules

**Event based rules**
- [Lexicon] Danger panel replaced with `form.error_summary` component on lexicon page
- [Lexicon] Table caption added to lexicon example rules table
- [Lexicon] Rules table updated with correct datatable responsive markup
- Remove link replaced with button with improved a11y label
- Rule block legend text updated to provide a better SR experience
- Error class moved from wrapping rule block `fieldset` to `.rule-block-wrapper` to keep error styling to the block and not bleed over unto And/Or buttons
- Rule block errors should now be passed to the individual fields and not passed via the `error` option on the block - this ensures error text is correctly associated to inputs and not orphaned. 
- Headings bumped to `h2` on rules blocks to avoid axe warnings

**Logic based rules**
- [Lexicon] Lexicon page added to rules patterns for logic based rules pattern
- [Lexicon] XFP rules page playground removed
- Form input markup updated to ensure labels are associated to inputs
- Rule blocks now use `fieldset` and `legend` rather than `div`'s to improve SR experience
- Rule block headings `span.rule-block__heading` changed to `h2`'s to improve SR navigation
- Rule block colour headings now have `font-weight: bold` to ensure coloured headings pass colour contrast - previously they did meet the wcag definition of large text but failed due to not being bold.

### Datatables
- Fixed an issue where collapsing datatables with `data-select="false"` would incorrectly show an empty first column (containing the hidden responsive toggle buttons).
- Fixed an issue where the first column was not sortable when `data-select="true"` was used.

Documentation update to event and logic examples to follow.

## Breaking change checklist

| ---- | ---- |
| require changes to `main.js` | ✅  No |
| require changes to `index.js` | ✅  No |
| require changes to `pulsar.scss` | ✅  No |
| require changes to `gruntfile.js` | ✅  No |
| require changes to `package.json` (not including dev dependencies) | ✅  No |
| require changes to `base.html.twig` (or other core views like header/footer) |  ✅  No |
| require new arguments to be passed to a js component | ✅  No |
| require markup changes to maintain functionality | 🔴  YES (Numerous changes to rules views) |
| require changes to existing unit tests | ✅  No|


jamesjacobs:
Fixes issues raised in #1180

stylesheets/_component.rules.scss
```diff
@@ -172,15 +178,32 @@
...
 .conditional-block--flush {
     margin: 24px 0 0 58px;
+    position: relative;
 
     @include respond-min($screen-tablet) {
         margin: 24px 0 0 68px;
     }
+
+    &::after {
+        color: #dedede;
```
Stanton, line 188: anything in the palette we can use?

stylesheets/_component.rules.scss
```diff
@@ -172,15 +178,32 @@
...
+        content: '\f061';
+        font-family: 'FontAwesome';
+        font-size: 19px;
```
Stanton, line 191: Will `$font-size-large` be enough?

stylesheets/_component.rules.scss
```diff
@@ -189,13 +212,27 @@
...
         top: 0;
         width: 4px;
         z-index: 1;
     }
+
+    .btn {
+        margin-left: 25px;
+    }
+
+    &::after {
```
Stanton, line 227: same var comments as above?

stylesheets/_component.rules.scss
```diff
@@ -493,6 +535,7 @@ td.table__form .rules + .form__group:last-of-type {
     }
 
     .rule-block__heading {
+        font-weight: bold;
```
Stanton, line 538: preferred over `$font-family-regular` ?

stylesheets/_component.rules.scss
jamesjacobs: Swapped

stylesheets/_component.rules.scss
jamesjacobs: Swapped

stylesheets/_component.rules.scss
jamesjacobs: Used to avoid contrast warnings in SI. Alternative would be to change logic bg colours.

PR #1284, opened 2020-08-03
Filterbar select2 placeholder and tab.js fixes

jamesjacobs:

### Filter bar select2 fixes
1. Fixed an issue where select2's in the filter bar that had a placeholder would be displayed on page load (see below)
screenshot_2020-07-27_at_18 01 05
2. Also fixed an issue where filterbar select2's with multiple selections would be displayed incorrectly - "label:Option1Option2" , rather than "label: Option1, Option2".

### Tab.js
#1258 introduced an issue where event handlers and JS would break on tab change. This resulted in the following elements being broken (after a js tab change): 

- select2
- colour picker
- date picker
- time picker
- repeater
- choice block radios and checkboxes
- dropzone
- jstree

This branch fixes the above issues and moves `tab.js` enhancements to their own class with relevant tests. As this functionality is not included in `tab.js`, the new `TabEnhancements` class file will need including and initialising in products using JS tabs.

### DIFF
--- a/js/TabEnhancements/TabEnhancements.js
+++ b/js/TabEnhancements/TabEnhancements.js
@@ -1,0 +1,66 @@
+var $ = require('jquery');
+
+/**
+ * Enhance tab.js behaviour
+*/
+class TabEnhancements {
+    /**
+     * Initialise
+     * @param {jQuery} $html - a jQuery wrapper of the html node
+     */
+    init ($html) {
+        this.$html = $html;
+
+        if(typeof $.fn.tab === 'undefined') {
+            console.warn('PULSAR: tab.js must be loaded in order to use TabEnhacements.js');
+        }
+
+        // Make sure tab panes are at least as high as the tab list (cms legacy tabs list)
+        this.$html.find('.tabs > .tabs__content > .tab__pane').css('min-height', this.$html.find('.tabs__list').height());
+
+        // Switch the first active tab to use <main> with the skip target
+        this.changeActiveTabElementToMain(this.$html.find('.tab__pane.is-active'));
+
+
+        // Make the current tabs .tab__content <main> and the previous tabs .tab__content <div>
+        // to avoid multiple <main>'s in the DOM at one time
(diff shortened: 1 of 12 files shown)


js/TabEnhancements/TabEnhancements.js
```diff
@@ -0,0 +1,66 @@
...
+        if(typeof $.fn.tab === 'undefined') {
+            console.warn('PULSAR: tab.js must be loaded in order to use TabEnhacements.js');
+        }
+
+        // Make sure tab panes are at least as high as the tab list (cms legacy tabs list)
+        this.$html.find('.tabs > .tabs__content > .tab__pane').css('min-height', this.$html.find('.tabs__list').height());
+
+        // Switch the first active tab to use <main> with the skip target
+        this.changeActiveTabElementToMain(this.$html.find('.tab__pane.is-active'));
+
```
Stanton, line 23: spacing

js/TabEnhancements/TabEnhancements.js
```diff
@@ -0,0 +1,66 @@
...
+
+        // Make the current tabs .tab__content <main> and the previous tabs .tab__content <div>
+        // to avoid multiple <main>'s in the DOM at one time
+        this.$html.find('.nav-inline [data-toggle="tab"]').on('show.bs.tab', (event) => {
+            const $activeTab = this.$html.find($(event.target).attr('href'));
```
Stanton, line 28: Is there any benefit to just using `event.target` and `event.relatedTarget` instead of fetching the href here? Smells like a lot of the following methods might be able to work just fine this way?

jamesjacobs: Not sure I follow, `event.target` and `event.relatedTarget` refer to the `a.nav-inline__link` of the new tab and the `a.nav-inline__link` of the old tab, not the tab itself. So we need to grab the href to find the `id` of the new/previous `div.tab__pane`.

Stanton: Ah ok, I thought it referred to the tab pane

jamesjacobs: 👍

PR #1291, opened 2020-08-18
Tab enhancement fixes

jamesjacobs:

1. The behaviour added in #1258 and #1284 that swapped the `main` element to a `div` and visa-versa on JS tab changes has been removed.
2. Markup in `tab.html.twig` includes `<main id="skip-target>` to ensure pages that don't use tab.js and reload the page on tab change have the skip target ID set on `<main>`.
3. The swapping of `id="skip-target"` is managed in `TabEnhancements.js`, so that when JS tabs are changed, the skip target ID moves over to the active tabs `main` (and is removed from non-active tabs main)

This does mean that we will see the following markup validation failure for pulsar pages which use JS tabs:
```
A document must not include more than one visible “main” element
```
However, the DOM is valid as other `main`'s outside of the currently active tab are hidden from the a11y tree via `display: none` on non-active `tab__pane`'s. The only way around that would be to add the `hidden` property to non active `main`'s  at a markup or twig level, when we know JS tabs are being used. Something to discuss.

### DIFF
--- a/js/TabEnhancements/TabEnhancements.js
+++ b/js/TabEnhancements/TabEnhancements.js
@@ -18,6 +18,9 @@ class TabEnhancements {
         // Make sure tab panes are at least as high as the tab list (cms legacy tabs list)
         this.$html.find('.tabs > .tabs__content > .tab__pane').css('min-height', this.$html.find('.tabs__list').height());
 
+        // Remove skip target IDs set in markup
+        this.$html.find('main.tab__content').removeAttr('id');
+
         // Add the skip-target ID to the active main
         this.$html.find('.tab__pane.is-active main.tab__content').attr('id', 'skip-target');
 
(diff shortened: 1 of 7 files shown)


jackbentley:
For the more than one visible main validation error message, you need to add the attribute `hidden` to all `main` elements which are not active.

The duplicate ID validation message is probably OK as this won't be seen once the JS initializes, and there shouldn't be duplicates for pages without

jamesjacobs:
As discussed with @jackbentley, markup validators validate the source markup and not the DOM so adding `hidden` doesn't resolve the issue. Going with the middle ground of `role="main"` rather than `<main>` to reduce validation errors on js tab pages until a better solution can be looked at. Duplicate ID error will still be present on JS tab pages.

PR #1297, opened 2020-08-24
Add tab header to sit above tab settings and tab content

Stanton:
Required for XFP Data Retention UI, adds a new block which sits above the tab sidebar & content, but below the actions bar.

Example:

```twig
{% block tab_header %}

  {{
      html.panel({
          'title': '',
          'body': form.toggle_switch({
            'label': 'Clear user data on submission'
          }) ~ '<p>If enabled, any system-wide data retention policies will not apply to this form. Instead, you’ll need to define new policies specific to this form.
Choosing to override global policies and not defining any form specific policies will mean form data is never removed. This is not recommended.</p>',
      })
  }}

{% endblock tab_header %}

{% block tab_settings %}

    <div class="settings-section settings-section--primary">
        <h2 class="settings-section__heading">Settings</h2>

        {{
          form.toggle_switch({
            'label': 'Clear user data on submission'
          })
        }}

        <p>Enabling this option will prevent all user-submitted data from being saved on this website. You must therefore ensure that at least one rule has been configured, in order to export the data to a third party system instead (e.g Jadu CXM).</p>
        <p>This option will also prevent users downloading a PDF of their submission. Should you wish, the user's email receipt can instead be configured to have this PDF attached.</p>
    </div>

{% endblock tab_settings %}
```


Stanton:
@james-manley Can you provide some detail about when you need this change, please? Presuming as part of `11.0.0beta2` ?

stylesheets/_layout.settings.scss
```diff
@@ -385,6 +385,7 @@
         align-items: flex-start;
         display: table;
         display: flex;
```
jamesjacobs, line 387: I know this hasn't changed in this PR but we have two `display`'s here

stylesheets/_layout.tabs.scss
```diff
@@ -232,3 +232,8 @@
         @include col-span-width(9);
     }
 }
+
+.tab__header {
+    flex-basis: 100%;
```
jamesjacobs, line 237: Feel like this should be set within `_layout.settings.scss` as that's the only place `.tab__header` would be within an element with `display: flex` (other layouts don't use flex), perhaps nested within the the rule block that contains the `flex-wrap` change above.

views/lexicon/patterns/tab-header/table.html.twig
```diff
@@ -0,0 +1,304 @@
...
+                'Modified': '2002-02-08 12:00',
+                'Live': true,
+                'Visible': true
+            }
+        ]
+    }
+%}
+
+{% block tab_header %}
+    <p>tab header goes here</p>
```
jamesjacobs, line 293: Could we add this to the settings panel too, with it having it's own settings specific flex styling. Would be helpful for future layout testing.

jamesjacobs (changes requested):

stylesheets/_layout.tabs.scss
```diff
@@ -232,3 +232,7 @@
         @include col-span-width(9);
     }
 }
+
+.tab__header {
+    width: 100%;
```
jamesjacobs, line 237: I'd move this up with flex-basis as we know it's always going to be a div and I suspect that the width is only needed when flexing?

Unless there's another reason why we need to specify it?

jamesjacobs (approved):
Approved with a minor point.

PR #1313, opened 2020-09-08
Ensure AA contrast for help/warning states within table forms

Stanton:
Fixes a AA issue reported in CMS, where validation states within a table__form container don't meet contrast.

**Before**
Screenshot 2020-09-08 at 12 25 52

**After**
Screenshot 2020-09-08 at 12 25 19

To be included in `11.0.0beta2`

### DIFF
--- a/stylesheets/_component.forms.scss
+++ b/stylesheets/_component.forms.scss
@@ -445,6 +445,16 @@ fieldset {
   }
 }
 
+.table__form .form__group:not(.form__group--partial-state) {
+    &.has-warning {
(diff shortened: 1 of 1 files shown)


jamesjacobs (commented):
Would it not be simpler and result in less CSS to lighten the BG of the table form? You might have already tried that?

Stanton:
Better solution, but more risk with unintended regressions

jamesjacobs (approved):
Approved but would really like to see us avoiding selectors with `.not()`'s, they always lead to pain further down the line due to super high specificity.

PR #1320, opened 2020-09-14
Make sure horizontal scrollbar is visible for tabbed navigation inside piano content

Stanton:
Related to CMS-6986

This rule causes the scrollbar to not be visible, the git history says that it was added to resolve an issue with full screen tables in the Perceptive Content integrations UI.

I’ve tested this UI with and without this ruleset active, in the usual browsers, and there is no negative effect with removing it, the presumption at this time is that there have been other changes to either Pianos or Tables which render this changeset redundant.

### DIFF
--- a/stylesheets/_component.piano.scss
+++ b/stylesheets/_component.piano.scss
@@ -329,15 +329,3 @@ $piano-bg: #fafafa;
 .setting__actions {
     padding: 10px;
 }
-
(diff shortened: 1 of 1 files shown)


jamesjacobs (approved):

Stanton:
Requested for `11.0.0beta3` by @theGreener

PR #1322, opened 2020-09-16
Set a consistent font size for card titles so that any heading tag can be used

Stanton:
Requiring cards to use a h3 means we might be using inappropriate heading levels. This change also allows non `h*` tags to be used.

**Before**
Screenshot 2020-09-16 at 14 03 38
**After**
Screenshot 2020-09-16 at 14 02 21

### DIFF
--- a/stylesheets/_component.cards.scss
+++ b/stylesheets/_component.cards.scss
@@ -27,6 +27,7 @@
 
 .card__title {
     font-family: $font-family-regular;
+    font-size: $font-size-large;
 }
 
 .card__body {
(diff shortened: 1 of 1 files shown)


jamesjacobs (commented):
Feels like we should also set `line-height` and `margin` which would be lost if non `h*` tags are used?

Stanton:
> Feels like we should also set `line-height` and `margin` which would be lost if non `h*` tags are used?

Agreed, added.

jamesjacobs (approved):

PR #1323, opened 2020-09-16
Stop flush inputs inheriting nowrap from horizontal table cells

Stanton:
Flush inputs are being affected by the `nowrap` rule which forces tables to maintain horizontal width and rely on scrolling. This change allows flush inputs to wrap and put the input on the left edge as expected.

**Example**
table-flush-inputs

### DIFF
--- a/stylesheets/_component.tables.scss
+++ b/stylesheets/_component.tables.scss
@@ -282,6 +282,11 @@ td.row__actions {
     th,
     td {
         white-space: nowrap;
+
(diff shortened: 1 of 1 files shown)


jamesjacobs (commented):
Looks good - before I hit approve, has this been browser tested?

Stanton:
> Looks good - before I hit approve, has this been browser tested?

It has, though it's all been through direct manipulation via devtools

jamesjacobs (approved):

PR #1326, opened 2020-09-21
Consistent sticky sidebar behaviour

jackbentley:
Update the behaviour of sticky sidebar to be consistent regardless of main content height.

Additionally fixes issues with "popping" of the sidebar when using the new `tab_header` block. 
And fixes spacing issues at the bottom of the settings panel when in "sticky" mode.

Once merged, we can remove our custom code from XFP which currently allows this behaviour.

**Before:**
scroll-before-emptyscroll-before-full

**After:**
scroll-after-emptyscroll-after-full


stylesheets/_component.calendar.scss
```diff
@@ -5,7 +5,6 @@
 }
 
 .ui-datepicker {
-    padding-bottom: 1em;
```
Stanton, line 8: Is this relevant to this change?

stylesheets/_layout.settings.scss
```diff
@@ -584,19 +585,26 @@
...
-    .tab__inner.is-sticky .tab__content {
-        left: calc(370px + 1%);
-        max-width: calc(100% - 370px);
-        position: relative;
-    }
+@include respond-min($screen-medium) {
+    .tab__inner.is-sticky {
+        min-height: 100vh;
+        padding-bottom: 0;
+        margin-bottom: -37px;
```
Stanton, line 592: Magic number needs explaining in a comment, presume it's to cater for footer?

Stanton (changes requested):
1. Updated js methods need test cases editing/adding (see codecov failure)
2. There needs to be some visual indication that the sidebar does not contain any items underneath the footer, some additional spacing is required.

sidebar

stylesheets/_component.calendar.scss
jackbentley: Yes and no. It causes additional padding at the bottom of every page (like the footer) which also causes additional scroll on the right-hand content as, unlike the footer, it isn't counteracted in sticky mode.

It looks like this CSS rule is overridden when the date picker shows by in-line styles. So it's essentially useless.

Screenshot 2020-09-22 at 17 37 48

stylesheets/_layout.settings.scss
jackbentley: It is indeed to cater for the footer height. Do we have a variable for this?

jackbentley:
> 1. Updated js methods need test cases editing/adding (see codecov failure)
> 2. There needs to be some visual indication that the sidebar does not contain any items underneath the footer, some additional spacing is required.

@Stanton Thanks for taking a look

1) Added a new test case
2) Can't replicate that. For me, the scrollbar is correct. This is what I meant by `And fixes spacing issues at the bottom of the settings panel when in "sticky" mode.`

Screenshot 2020-09-22 at 17 52 34

stylesheets/_layout.settings.scss
Stanton: Have added (and wired in) a $footer-height token which does the same calculation of line height + padding + border

Stanton (approved):

Stanton (approved):

PR #1331, opened 2020-09-28
Stop extracting text from a DOM node and interpreting as HTML

Stanton:
These changes fix a vulnerability where strings are extracted from the DOM interpreted as HTML. The suggested fix for these from CodeQL was to wrap the values in `$.find()` which will interpret the value as CSS rather than raw HTML.

More info (if you have access) https://github.com/jadu/pulsar/security/code-scanning/15?query=ref%3Arefs%2Fheads%2Fdevelop

If there are any other preferred/more performant methods of patching these holes, or if the use of `find` here is sufficient, please let me know!

Results of these changes: https://github.com/jadu/pulsar/pull/1331/checks?check_run_id=1220879081


mikedevelops (commented):
If we expect there to be more instances of this, it might be nice to abstract the escaping process so we can express how that happens in 1 place. 

I don't know much about this `$.find` workaround, but should we want to replace that with proper html escaping/sanitizing in the future, it may be worth having a `Pulsar::escapeHTML` helper that we can change in one place as appose to making a change in all instances we are using text from the DOM.

Stanton:
> If we expect there to be more instances of this, it might be nice to abstract the escaping process so we can express how that happens in 1 place.

There are a lot of instances in DropZone & FilterBar, so indeed a common escape method may be more desirable especially with the 12.0 release queued

Stanton:
Or now we're using jQuery3, https://api.jquery.com/jQuery.parseHTML/ may be the more appropriate method instead of find...

mikedevelops:
ParseHTML may not be enough:

> Most jQuery APIs that accept HTML strings will run scripts that are included in the HTML. jQuery.parseHTML does not run scripts in the parsed HTML unless keepScripts is explicitly true. However, it is still possible in most environments to execute scripts indirectly, for example via the  attribute. The caller should be aware of this and guard against it by cleaning or escaping any untrusted inputs from sources such as the URL or cookies. For future compatibility, callers should not depend on the ability to run any script content when keepScripts is unspecified or false.

Stanton:
Results of this change https://github.com/jadu/pulsar/pull/1331/checks?check_run_id=1220879081

jamesjacobs (approved):

PR #1339, opened 2020-11-09
Beta/12.0.0beta1

jamesjacobs:
This branch contains the following:

1. Fixed an issue where the help button would disappear
2. Fixed an issue when dropdown list dividers were bigger than they should be
3. Fixed an IE11 issue where layout would shift under the main nav
4. Fixed dropdown danger link focus styles
5. Updated the Modal focus service to allow for edges cases where none of the standard modal buttons/attributes are in use
6. Updated the Man Nav component so that the it closes correctly when elements in the toolbar or footer are clicked or ESC is pressed and focus is now managed correctly.
7. Updated the mobile help button - this now displays as a button, rather than an icon and prepends itself to the top of `tab__content` - this is to overcome an issue where in certain layouts the help button was not added to the DOM.
8. Updated the TooltipListener to better allow for content with tooltips loaded via AJAX
9. Updated the skip to content link markup, it's now promoted to a landmark with `role="complementary" aria-label="skip to main content"`
10. Removed the `aria-label` from the main branding navigation link which was cause a11y tool errors.
11. Removed some old CXM specific styles

**Note: mention z-index changes in release notes.**
`$zindex-dropdown` changed from `1060` to `1065`
`$zindex-flash` changed from `1065` to `1065`
This is to fix an issue where flash messages incorrectly appeared over dropdowns (like user settings dropdown).


js/NavMainComponent.js
```diff
@@ -126,18 +129,39 @@ NavMainComponent.prototype.init = function () {
...
 
+/**
+ * Check if any navs are open
+ * @returns {boolean}
+ */
+NavMainComponent.prototype.isNavOpen = function () {
+    var component = this,
+        isMobile = !component.window.matchMedia('(min-width: 992px)').matches;
+
+    if (component.$navQuaternary.hasClass('is-open')) {
```
Stanton, line 152: It's a minor thing, but seems like you've ordered this from least-likely to most-likely, flipping it would mean we could get out of this conditional sooner in most cases

js/Tooltips/TooltipListener.js
```diff
@@ -57,6 +58,49 @@ class TooltipListener {
...
+    isInstantiated(element) {
+        for (const tippy of this.tippys) {
+            if (tippy.reference === element) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+
```
Stanton, line 89: remove newline

stylesheets/_config.variables.scss
```diff
@@ -220,8 +220,8 @@ $zindex-tour:                         1000 !default;
 $zindex-footer:                       1020 !default;
 $zindex-tooltip:                      1055 !default;
 $zindex-sticky:                       1060 !default;
-$zindex-dropdown:                     1060 !default;
-$zindex-flash:                        1065 !default;
+$zindex-flash:                        1064 !default;
```
Stanton, line 223: This needs a special mention in the release notes, we forever have z-index regressions in CMS so this will need specific testing

views/pulsar/v2/helpers/nav.html.twig
```diff
@@ -131,11 +131,11 @@
             )
         }}
     >
-        <{{ attributes(options|only('aria-label href data-toggle data-target'), { 'tag': tag }) }} 
-            
-            class="nav-link t-nav-link" 
-            {% if options.class is defined and 'is-active' in options.class %} aria-current="page" {% endif %} 
-            {% if options.href starts with '#' or tag == 'button' %} aria-haspopup="true" aria-expanded="false" 
+        <{{ attributes(options|only('aria-label href data-toggle data-target'), { 'tag': tag }) }}
+
```
Stanton, line 135: I don't think this newline is intentional...

Stanton (changes requested):

stylesheets/_config.variables.scss
jamesjacobs: 👍

js/NavMainComponent.js
jamesjacobs: Done

Stanton (approved):

PR #1351, opened 2021-01-06
Switch grunt sass to use dart-sass

### LINKED ISSUE #1338: LibSass deprecated
https://sass-lang.com/blog/libsass-is-deprecated

Stanton:
Libsass has been deprecated, this change makes the `grunt sass` command use dart-sass instead of node-sass.

1. Adds `dart-sass` and `fibers` package to devDependencies
2. Update gruntfile to work with dart-sass options
2. Update files which use string interpolation like `#{$state}` to use `#{'' + $state}` (only an issue for Pulsar colour palette values only)

Number 3 may represent a breaking change that can be ironed out during smoke testing, I've searched CMS and XFP and there's no custom code using state modifier classes.

Closes #1338


jamesjacobs (approved):
There are 3 instances of string interpolation in CXM. I've quickly tested these with the `'' + $var` change and there were no issues.

Stanton:
I'm going to merge this one as we already need to make markup changes in products for the upcoming release to address issues in rules and datatables. This change will also hopefully banish the need for `npm rebuild node-sass`.

PR #1361, opened 2021-01-08
Table caption styling

### LINKED ISSUE #1347: Add styling for table captions
Currently table captions are either accessibly hidden with `hide` or not used. Add styling so that if table captions are displayed then we have uniform styling rather than relying on browser stylesheet defaults.

Ensure new styling covers the use of heading tags within the caption (aids SR/AT navigation).

jamesjacobs:
This branch adds basic styling to table `<caption>` elements.

Before (chrome default styling):
Screenshot 2021-01-08 at 15 30 39

After:
Screenshot 2021-01-08 at 15 30 48

Fixes #1347


Stanton:
Changelog needs updating, might as well do that before I review so it doesn't need re-review

jamesjacobs:
@Stanton Updated changelog 👍

Stanton (dismissed):

PR #1363, opened 2021-01-11
Added scope=“col” to repeater and datatable th’s

### LINKED ISSUE #1343: Table THs should use scope="col"
Table headers (`thead th`'s) should use the `scope="col"` attribute to help user of assistive technology identify columns and potentially the difference between row headers and column headers.

Table and datatable helpers to handle where possible.

Related links:
https://www.w3.org/WAI/tutorials/tables/one-header/
https://www.w3.org/TR/WCAG20-TECHS/H63.html

Documentation examples will need to include and cover this.

jamesjacobs:
Fixes #1343 

This branch contains changes to both the `html.datatable` and `form.repeater` helper markup. As detailed in the issue above, table column headers should include the correct scope, especially where the table data could be classed as ambiguous or where the first column may not contain a header relevant to the table data.

See https://www.w3.org/WAI/tutorials/tables/one-header/ and https://www.w3.org/TR/WCAG20-TECHS/H63.html


views/pulsar/v2/helpers/html.html.twig
```diff
@@ -1581,13 +1581,13 @@ Datatable - https://jadu.github.io/pulsar/components/datatable/
                     {% for column in options.data.columns %}
                         {% if loop.first %}
                             {% if options['data-select'] is not defined or (options['data-select'] is defined and options['data-select'] != 'false') %}
-                                <th class="table-selection" data-orderable="false">
+                                <th class="table-selection" data-orderable="false" scope="col">
```
Stanton, line 1584: Really nitpick, but can this be at the start like the other helper, or at least before the data attribute?

Stanton (approved):

PR #1366, opened 2021-01-13
Fix form compound mobile spacing

### LINKED ISSUE #1107: Improve mobile spacing of form__group--compound and input-groups
There's inconsistent spacing when using `fieldset class="form__group form__group--compound"` and `input-group` at mobile VP widths (see screenshot).

Issues:
* `legend` with `control__label` has incorrect padding/margin vs standard `label`'s
* `input` top margin needs removing so `input` vertically aligns with appended button

spacing-issues

jamesjacobs:
This branch addresses issues with the mobile spacing of `form.compound`.

**Before (note the gap between the main label and the missing margin on the last input):**
Screenshot 2021-01-13 at 16 02 43

**After:**
Screenshot 2021-01-13 at 16 06 59

Note that a right margin on the last `.form__control` does exist on desktop, though that shouldn't cause any issues (tested in lexicon). The previous code tried to remove that but failed due to the last input in the compound not being the last child if help or error text was included.

Fixes #1107 (partially - as the input group issue raised isn't a supported pattern and therefore would require more investigation and/or a product scratch fix / alternative).

Browser tested in all supported browsers.


stylesheets/_component.forms.scss
```diff
@@ -756,6 +756,12 @@ fieldset.form__group {
 }
 
 .form__group--compound {
+    > legend.control__label {
```
Stanton, line 759: Why not just `> .control__label`? this shouldn't target the control labels inside `controls` if that was the concern?

jamesjacobs: Valid point, removed 🔥

PR #1375, opened 2021-02-09
First attempt to refactor button stylesheet

Stanton:
Target branch = `purple`

This is a first-pass refactor of the button component which aims to simplify the styles required to generate Pulsar's buttons and reduce overall specificity.

It is expected that there will be additional work required before this PR is acceptable.

Notable changes:
* Bottom outset border has been removed
* All reliance on `!important` has been removed
* Font size increased to 18px
* Button target height increased to 38px to match planned input sizes
* Outline button borders are now 1px instead of 2px

Other notes:
* Table link styles have been stripped out in this change, fully expect additional work to happen for table links as part of Purple
* I plan on removing the `info` blue style due to lack of adoption, but as that doesn't just affect buttons I'll do this as a separate PR

### Specificity change
button-after

### Screenshots
Screenshot 2021-02-09 at 12 57 37
Screenshot 2021-02-09 at 12 57 43


stylesheets/_component.buttons.scss
```diff
@@ -1,385 +1,110 @@
...
- * 4. Normalize `line-height`. Cannot be changed from `normal` in Firefox 4+.
- * 5. Prevent button text from being selectable.
- * 6. Prevent wrapping across multiple lines.
- * 7. Remove excess padding and border in Firefox 4+
- */
-
-$btn-padding-top:           10px;
-$btn-padding-bottom:        7px;
-$btn-padding-x:             14px;
+// Variables
```
jamesjacobs, line 1: Move to `_config.variables.scss`?

stylesheets/_component.buttons.scss
```diff
@@ -1,385 +1,110 @@
...
-    &.is-disabled,
-    &.disabled {
-        cursor: not-allowed !important;
+        pointer-events: none;
     }
 }
 
+// State variations
 @each $state, $state-color, $state-color-alt in $state-colors {
     .btn--#{'' + $state} {
```
jamesjacobs, line 64: Wrap in `@if $state != 'base' {}` to avoid unused `--base` styles being outputted in compiled CSS.

stylesheets/_component.buttons.scss
```diff
@@ -1,385 +1,110 @@
...
-    }
-
-    &.btn--naked {
-        padding: $btn-padding-top $btn-padding-x $btn-padding-bottom !important;
-    }
-
-    &.btn--small {
-        padding: ($btn-small-padding-top - 2) ($btn-small-padding-x - 2) ($btn-small-padding-bottom - 2) !important;
+    background: none;
+    border: 1px solid color(base);
```
jamesjacobs, line 72: See notes on this. Feels like 1px is too thin.

stylesheets/_component.buttons.scss
```diff
@@ -1,385 +1,110 @@
...
+    }
+}
+
+// Button Constructor
 .btn {
-    @include inline-block;
-    -webkit-appearance: none; /* 1 */
-    background: transparent;
+    @include button-variant('base');
     border: 0;
```
jamesjacobs, line 30: See notes on button borders.

stylesheets/_component.buttons.scss
```diff
@@ -1,385 +1,110 @@
...
-        }
+    // Double chain focus states to stop outline state variation overriding background colour
+    &.btn--outline:focus,
+    &.btn--outline:hover:focus {
+        @include pulsar-button-focused;
+        border: 0;
     }
 
     @each $state, $state-color, $state-color-alt in $state-colors {
         &.btn--#{'' + $state} {
```
jamesjacobs, line 83: Wrap in `@if $state != 'base'` to avoid outputting unused `--base` styles in compiled CSS.

stylesheets/_component.tables.scss
```diff
@@ -17,33 +17,33 @@
...
-        &:focus {
-            border: 0;
-            color: color(black);
-        }
-    }
-
-    [href].btn {
-        border-bottom: 0;
-    }
+    // [href],
```
jamesjacobs, line 20: Assume we're looking to replace these styles with a class?

jamesjacobs (commented):
### Notes & thoughts so far

**Standard buttons**
- Default grey button without a border or outline may look a little lost on a grey piano or reports UI. While not a WCAG requirement, it's recommended that component boundaries  are delineated  (see boundaries https://www.w3.org/WAI/WCAG21/Understanding/non-text-contrast.html)

**Outline buttons**
- Slight visual shift on focus
- Border size reduction? Harder to see , especially on coloured backgrounds such as panels.

**Compiled CSS**
- `btn--base` styles are still present, contents of each loop needs wrapping in `@if $state != 'base'`
- All button styles are duplicated due to `@import 'component.buttons'` in `_component.remove-button.scss` - removing that @import doesn't seem to cause any issues in lexicon - needs further testing.
- Duplication of styles and `--base` styles in compiled CSS also present in label CSS. Might be out of scope for this PR but potentially worth fixing at the same time.

**Questions:**
- PR mentions increasing height to 38px - should we consider bumping this to 44px to meet AAA on 2.5.5 Target size (https://www.w3.org/WAI/WCAG21/Understanding/target-size)

stylesheets/_component.tables.scss
Stanton: Maaaybe, it's one to discuss for sure

stylesheets/_component.buttons.scss
Stanton: I'm going to run with it for now, most notable design systems use 1px for outline buttons and I'm hoping this will reduce our reliance on naked buttons a little by being subtle enough to not draw too much focus

stylesheets/_component.buttons.scss
Stanton: Will be moving when things stabilise

Stanton:
> * Default grey button without a border or outline may look a little lost on a grey piano or reports UI. While not a WCAG requirement, it's recommended that component boundaries  are delineated  (see boundaries https://www.w3.org/WAI/WCAG21/Understanding/non-text-contrast.html)
Changed default outline button to be same as inverse for the time being

> * Slight visual shift on focus
Will tweak

> * Border size reduction? Harder to see , especially on coloured backgrounds such as panels.
Will cross those bridges as they come up, might be that most of the time we shouldn't/wouldn't want to use an outline button on a non-white background, panels/flash might be exceptions though

> * `btn--base` styles are still present, contents of each loop needs wrapping in `@if $state != 'base'`
Good shout, added

> * All button styles are duplicated due to `@import 'component.buttons'` in `_component.remove-button.scss` - removing that @import doesn't seem to cause any issues in lexicon - needs further testing.
Will remove that @import as part of this PR, will watch out for any oddities, looks like it was a brute force specificity override in the past

> * Duplication of styles and `--base` styles in compiled CSS also present in label CSS. Might be out of scope for this PR but potentially worth fixing at the same time.
Not touching labels here

> * PR mentions increasing height to 38px - should we consider bumping this to 44px to meet AAA on 2.5.5 Target size (https://www.w3.org/WAI/WCAG21/Understanding/target-size)
44 is massive, legit massive, I can't bring myself to do it, we're not intending to meet AAA for anything else so I can't really justify the amount of work it would take to bring all other components up to this minimum size

stylesheets/_component.buttons.scss
Stanton: Leaving for now

Stanton:
Replaced by #1380

PR #1380, opened 2021-02-19
Button and label refactor

Stanton:
I initially wanted to keep label and button refactor separate, however they're so closely related and we have buttons inside labels so it's not really possible to test the label change without depending on the buttons changes.

Includes all changes in #1375

Screenshot 2021-02-19 at 08 31 37

Notable changes:
* Labels are larger
* Text size increased to 14px (we may want to go whole hog and bump to 16?)
* All reliance on `!important` has been removed (smoke test needed to test the impact of this)
* `label--large` variant has been dropped (new label size is nearly the size of the current large variant)
* Labels inside links now retain the link underline to try make them more obviously links

Other notes:
* `@import` of labels component removed from toolbars scss to remove duplication
* Started adding some em based padding variables, would be nice to address this across the system but the change footprint will be too large
* Label variant mixin has been moved inside the label stylesheet, this matches buttons, and separating them out felt pointless as they're only used by this stylesheet

Questions:
* Custom label styles, do we use them in production anywhere? They'd need me to add immutability back to the focus styles, which isn't _necessarily_ an issue, but trying to remove !important rather than adding more


stylesheets/_component.tables.scss
```diff
@@ -17,33 +17,33 @@
...
-        &:focus {
-            border: 0;
-            color: color(black);
-        }
-    }
-
-    [href].btn {
-        border-bottom: 0;
-    }
+    // [href],
```
jamesjacobs, line 20: Plan for these?

stylesheets/_component.toolbar.scss
```diff
@@ -30,43 +28,56 @@
...
+        vertical-align: middle;
 
-        > .caret {
-            border-top-color: color(text);
+        @include respond-min($screen-desktop) {
+            font-size: 14px;
+            vertical-align: baseline;
         }
     }
+    // .btn:not(.btn--naked):not(.remove-button):not(.edit-button):not(.remove-button):last-of-type {
```
jamesjacobs, line 45: plan for these styles?

stylesheets/_component.toolbar.scss
```diff
@@ -109,16 +120,16 @@
     }
 }
 
-.toolbar-actions .toolbar-action--icon .btn:not(.btn--naked):not(.remove-button):not(.edit-button):not(.remove-button):last-of-type {
-    line-height: 16px;
-    padding-bottom: 6px;
-    padding-top: 6px;
+// .toolbar-actions .toolbar-action--icon .btn:not(.btn--naked):not(.remove-button):not(.edit-button):not(.remove-button):last-of-type {
+//     line-height: 16px;
```
jamesjacobs, line 124: and these?

stylesheets/_mixin.focus.scss
```diff
@@ -125,20 +125,19 @@
...
 @mixin pulsar-label-remove-button-focused {
-    background-color: color(background, focus) !important;
-    color: color(black) !important;
-    outline: none;
-
-    [class^="icon-"] {
-        color: color(black) !important;
-    }
+    box-shadow: inset 0 0 0 2px color(black);
+    // background-color: color(background, focus) !important;
```
jamesjacobs, line 136: plan for these?

jamesjacobs (changes requested):
Great work on reducing complexity and specificity 👏

I've taken this branch for a spin locally (not tested in CXM - just lexicon) and found the following issues:

**Buttons**
1. drop up button menu appear at the top of the screen and not directly above the button
2. remove button in labels looks slightly misaligned (vertically) - stands out more on focus
3. badges within button are now grey by default - assume this was intentional? 

**Labels**
1. SCSS for badges inside labels still outputs `--base` class

**Misc**
1. Colour picker button needs sizing adjustment
2. Dropzone "Browse files" button needs font size bump
3. Appended buttons / icon buttons on inputs need a size adjustment
4. Table detail close button focus styles cut off
5. Modal close X button needs a sizing adjustment - see on focus
6. Full screen modal buttons (download and close) need some tweaks on hover, font size and focus.
7. Mobile menu button still has old border radius

PR #1381, opened 2021-02-22
Fix infinite loop in 'more' responsive navigation menu

### LINKED ISSUE #1379: Resizing the window below a certain size, causes the window to become unresponsive
Open a page from CMS 20, and then resize it to be very small. When attempting to resize it back the window has become unresponsive and in Chrome will report `RESULT_CODE_HUNG`. This can be done when the full side menu is present in the window, and when it's been collapsed down to the "MENU" button already also (as pictured below).

**To Reproduce**
1. Go to a page in CMS 20.x.x (not replicable in CMS 19.x.x) and resize the widow vertically to be very small
2. Wait a second, and then try to resize it back to normal

**Expected behavior**
The window should be resizable back to what it was

**Screenshots**
VirtualBox_Windows 10_15_02_2021_15_02_48
VirtualBox_Windows 10_15_02_2021_15_03_08
VirtualBox_Windows 10_15_02_2021_15_03_25

**Desktop:**
 - OS: Windows 10
 - Browser: Chrome, Edge, IE 11
 - Version:  88.0.4324.150 (Chrome 64-bit), 88.0.705.68 (Edge 64-bit)

**Additional context**
This was an issue found by a user when trying to demonstrate scrolling not working on an XFP page (not reproducible at present)

Stanton:
The `hideNavigationItems` method goes into an infinite loop when all menu items are already hidden, this change keeps track of the number of hidden items every time `adjustNavItems` is called and breaks the loop if all items are hidden when the viewport is being resized.

Fixes #1379

**Before**
before

**After**
after


jamesjacobs (changes requested):
Missing related tests.

PR #1385, opened 2021-03-04
Address usability issue where the bottom item in a piano list is often missed when scrolling

Stanton:
This variable height padding adds a bit of extra runway to make sure the bottom of the piano list or content is always possible to be scrolled into view.

**Before**
piano-before
**After**
piano-after


jackbentley:
Looks like the scroll bar disappears behind the footer in the after screenshot?

Stanton:
> Looks like the scroll bar disappears behind the footer in the after screenshot?

Also present in the before.

jamesjacobs (approved):

PR #1403, opened 2021-07-06
Update breadcrumb markup scheme to match a11y guidelines

Stanton:
* Render breadcrumbs with an ordered list
* Remove hardcoded items separators and move that to CSS

https://www.w3.org/TR/wai-aria-practices-1.1/examples/breadcrumb/index.html


Stanton:
Included in #1412

PR #1404, opened 2021-07-07
Add novalidate attribute to form create helper

Stanton:
This change adds `novalidate` to the list of default attributes supplied by the `form.create` helper.

This allows products to fall back to server-side validation as per accessibility guidelines.


Stanton:
@jamesjacobs Do we still want this change?

PR #1481, opened 2023-01-31
Make sure scrollable nav containers are keyboard accessible

### LINKED ISSUE #1480: SiteImprove reporting issue with main navigation keyboard accessibility
**Describe the bug**
The current version of the SiteImprove accessibility checker reports an issue relating to the main navigation component. We are confident that the component's level of accessibility meets the guidelines, however due to the persistent nature of this component I recommend we take remedial action to prevent SI from reporting this issue, likely by the management of zero and negative tabindexes on hidden panels.

Failure to address this will result in repeated queries from QA teams and potentially start to be flagged by cypress.

Initially observed in Jadu Connect, verified in Pulsar Lexicon.

Jira Ticket: QU-5351

**Expected behavior**
The SiteImprove tool should no longer report this issue.

**Screenshots**
Screenshot 2023-01-30 at 09 23 02

Stanton:
Stops siteimprove plugin from flagging this as an issue.

Adds a tabindex on the flyout containers which are scrollable and so will now receive focus. A new focus outline has been added to maintain consistency with established focus states.

Screenshot 2023-01-31 at 11 49 19

Closes #1480

### DIFF
--- a/js/NavMainComponent.js
+++ b/js/NavMainComponent.js
@@ -325,9 +325,7 @@ NavMainComponent.prototype.closeSecondaryNav = function (action) {
 NavMainComponent.prototype.closeTertiaryNav = function () {
     var component = this;
 
-    component.$navTertiary.removeClass('is-open')
-        .attr('aria-hidden', 'true');
-
+    component.$navTertiary.removeClass('is-open').attr('aria-hidden', 'true');
     component.$navTertiary.find('.nav-list').removeClass('is-active');
 
     component.$navMain.find('[aria-expanded=true]')
--- a/js/NavMainComponent.js
+++ b/js/NavMainComponent.js
@@ -223,6 +223,9 @@ NavMainComponent.prototype.openSecondaryNav = function ($triggeringElement, even
     component.$navPrimary.find('.is-active').removeClass('is-active');
     component.$navPrimary.find('[href="' + target + '"], [data-target="' + target + '"]').addClass('is-active');
 
+    // Manage scrollable attributes
+    component.$navSecondary.attr('aria-hidden', 'false');
+
     // Manage focus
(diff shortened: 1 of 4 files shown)


jamesjacobs (changes requested):
As suggested, Site Improve is reporting a false positive here as the navigation elements, when open, have focusable elements which allow users to navigate through the list without content being hidden by scroll. 

I suspect SI is using this ACT rule: https://www.w3.org/WAI/standards-guidelines/act/rules/0ssw9k/proposed/ which states either the element is included in the sequential focus navigation or has a descendant in the sequential focus navigation - which in this case, it has, in the form of links and buttons. SI is potentially flagging this as the navigation elements are still present in the accessibility tree when closed while the links are not (due to `display: none`).

### **Issues found while testing this change**
**Hidden tab stops**
When the navs are closed, they are still present in the focus order (resulting in hidden tab stops). This is due to `tabindex="0"` being present on `.nav-secondary`, `.nav-tertiary` and `.nav-quaternary`.  The `tabindex` should be removed when the navs are closed. This issue can be replicated by KB tabbing. SI also reports this as "Hidden element has focusable content".

image

**Screen reader experience**
With this change, when the nav element gains focus, screen-readers read out all the contents of the focussed nav element, which creates unnecessary screenreader "noise".
To fix that, I'd recommend we add `aria-labelledby="NAV-HEADING-ID"` to each nav, which will result in the nav heading being read rather than a potentially long list of links on focus.

---

My preference would be to avoid this fix completely as it adds unnecessary tab stops for users in an effort to silence a SI false positive.

Stanton:
Decided to agree with James, we will not make a change here to address a false positive issue with the SiteImprove tool, any change will be non-trivial and add complexity and technical debt that doesn't benefit the design system, and decreases the real-world accessible experience of this component.

PR #1556, opened 2024-09-05
Qu 4501

Sectimus:
The file input does not retain it's files prior to submit, every time you attach a file, it was overwritten; Which isn't a problem if you are serializing the data yourself, but is a problem if you need to trust that your file input contains your file data.

## Before
image

> Generic binary data (or binary data whose true type is unknown) is application/octet-stream.
> - https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types

## After
image


Stanton:
Dropzone test suite failing so can't be merged, the Node14 environment falls over on github but returns the same failing tests when run locally.

Stanton (approved):

PR #1646, opened 2026-06-03
QU-6778 Upgrading twig/twig to ^3.27

irinikp:


composer.json
```diff
@@ -8,7 +8,7 @@
         }
     },
     "require": {
-        "twig/twig": "^2.0"
+        "twig/twig": "^3.27"
```
christopherp309, line 11: Please could we add `jadu/twig-style` so we have a twig cs fixer enforcing the Jadu code style on this repo? It would also be good to check if this works on twig 3 as historically we've only been using it on twig 2 repos.

https://github.com/jadu/twig-style

irinikp: I'm happy to add this, however there's currently a compatibility issue with `jadu/twig-style`.

It depends on `vincentlanglet/twig-cs-fixer: 2.0`, which is not compatible with projects using `twig/twig ^3`. As a result, running the lint step fails with runtime errors (e.g. missing methods like `getUnaryOperators()`).

This is a known issue and has been open for some time:  https://github.com/jadu/twig-style/issues/5

Until `jadu/twig-style` updates its dependency (or switches to a maintained alternative), integrating it would break the pipeline for any project on Twig 3.

Happy to revisit once that dependency is updated @christopherp309

irinikp: In the meantime, I've created a separate PR to introduce this change:  
https://github.com/jadu/pulsar/pull/1647

If the upstream issue gets resolved in the future, this branch can be picked up and completed without starting from scratch

christopherp309: Thanks Irini. It looks like a mitigation for this was developed here https://github.com/VincentLanglet/Twig-CS-Fixer/pull/362/changes but as you said, only released in `vincentlanglet/twig-cs-fixer: 3.x`

We won't be able to get jadu/twig-style to work with `vincentlanglet/twig-cs-fixer: 3.x` quickly because this version does not work with Twig 2, and we are blocked from using Twig 3 ourselves.

Therefore:
- In the short term, we can add a conflict with > twig 3.21 to jadu/twig-style. This won't fix the issue with installing it on Pulsar but will provide a partial resolution to https://github.com/jadu/twig-style/issues/5
- In the longer term, upgrade jadu/twig-style to require `vincentlanglet/twig-cs-fixer: ^3.0` so we are compatible with all Twig 3 versions.

irinikp: I agree that adding a conflict makes sense as a short-term safeguard, so that we don't end up with more people trying to install jadu/twig-style in Twig 3 projects where it's not compatible.
That said, I don't think we should consider this a resolution of the issue, but rather a temporary mitigation. It avoids the immediate breakage, but the underlying compatibility problem with twig-cs-fixer still remains.
Longer term, we'll still need to align on supporting twig-cs-fixer 3.x (and therefore Twig 3) properly.
Happy to contribute feedback when you come back to the upgrade

christopherp309: Yes, agreed that it's only a mitigation and the longer term solution of adopting twig-cs-fixer 3.x and Twig 3 is required. Until that's done it won't be possible to add jadu/twig-style to this repo as it will conflict with the security updates in the latest version of Twig as you've added here

christopherp309: @irinikp please could you try `jadu/twig-style: 2.0.0-alpha.1`? This version depends on `vincentlanglet/twig-cs-fixer: 4.0` and so is compatible with `twig/twig ^3 | ^4`

irinikp:
@christopherp309 am I allowed to merge this and tag? If yes, should I tag 16.0.0?

christopherp309:
> @christopherp309 am I allowed to merge this and tag? If yes, should I tag 16.0.0?

@irinikp decisions around tagging Pulsar are not up to me. Having said that, the twig major version bump indicates this should be a major version change in Pulsar to me.

Additionally, if you still wish to try incorporating jadu/twig-style I have just tagged `2.0.0-alpha.2` which is compatible with Twig ^3.15. Any testing you can do around this would be appreciated ahead of a full release.

